PR #15, opened 2018-02-26
Get a steel thread going on the platform tests

ryana:
Some context for what is in this PR:

- A steel thread of tests passing on the Listings.sol copied from demo-dapp
- A proposed directory structure modeled on how RequestNetwork does it (this slightly complicates CI tests so we may want to move away from it)

I'm not expecting this to get merged straight in. Figured it could serve as a good discussion catalyst.

### DIFF
--- a/packages/contracts/README.md
+++ b/packages/contracts/README.md
@@ -1,0 +1,24 @@
+# OriginProtocol Smart Contracts Documentation
+
+## Introduction
+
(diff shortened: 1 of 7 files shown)


ryana:
I'd like to hookup Codeship to this repo btw. I'm working on figuring out how to run it off my own fork (running w/ testrpc is slightly non-trivial). Once successful, an admin on the OriginProtocol org would need to go setup an account as well.

PR #24, opened 2018-03-10
Use develop as main dev branch

aiham:
Merging master into develop


wanderingstan:
Thanks @aiham ! 

Any tips on how to message people in the future that they should be making PR's to `develop` and not `master`?

aiham:
@wanderingstan from https://stackoverflow.com/a/14858326

> make develop the default branch that everyone sees when visiting the project. The downside is that anyone who clones it will get an unstable branch by default but all pull requests will go to the develop branch by default too.

aiham:
Alternatively you could flip it and use `master` and `stable` where all dev happens on master

aiham:
I think either way you'll need the default branch to be unstable to receive PRs to it automatically

PR #25, opened 2018-03-10
Use async/await in OriginService#submitListing

aiham:
- A lot cleaner than using raw promises (even though it's still promises under the hood)
- Also got rid of some tab characters


wanderingstan:
Awesome. We need to migrate all of our ugly promise()/resolve() code to `await`. So much cleaner.

PR #26, opened 2018-03-10
Add troubleshooting steps to READMEs

aiham:


wanderingstan:
Interesting, the `is not a contract address` error also comes up when doing a `migrate` (instead of `test`) and can be fixed by doing `migrate --reset`. I wonder if there's a `test --reset` option? 
```
+Error: Attempting to run transaction which calls a contract function, but recipient address 0x8cdaf0cd259887258bc23a92c0a6da92698644c0 is not a contract address
```

aiham:
@wanderingstan Yeah I didn't know about `--reset` until you mentioned it yesterday. I'll update these instructions with `--reset` instead.

PR #30, opened 2018-03-10
Add in-browser tests for JS modules

aiham:
Instructions in README.
Still need to flesh out all the tests.
There are some challenges in testing cases which require user interaction (such as accepting MetaMask dialogue).
These will probably have to be mocked eventually, probably using two sets of tests that do and don't have mocks.


wanderingstan:
This is a beautiful site! @aiham for the win. 
image

wanderingstan:
@ryana : @aiham and I talked last night about changes that will allow us to test wallet stuff without metamask. (Basically changing the web3 object to a stub that can be swapped out.) Once that's done, we can add these tests to codeship and really up our testing game. :)

PR #31, opened 2018-03-12
User-Registry-Service for Civic

that1guy:
Adding `user-registry-service.get()` and `user-registry-service.create()`.  Temporarily modified UserRegistry.sol - need to talk with team about how to store data in contract.

Need to talk about how we could auto-login the user via their wallet ID (if their wallet ID is already in our user-registry.sol contract). (https://github.com/OriginProtocol/platform/pull/31/files#diff-cf184f361bf7a42627ca98cccbd41d76R42)

Need to talk about how we're going to store Civic's decrypted user data in our user-registry.  I'm still fuzzy here. (https://github.com/OriginProtocol/platform/pull/31/files#diff-cf184f361bf7a42627ca98cccbd41d76R20)


wanderingstan:
Awesome! I will look at this today.

wanderingstan:
Amazing Brad. I'm *this* close to having it all working along with all the other merged PRs on platform.

I can see the first half of the flow working when I use the `origin` package from npm -- but it doesn't have userRegistryService so it craps out at the end. 

So I run `npm link @originprotocol/origin` to link to my local, updated platform. 

But then when I run, I get this error in Chrome console:
```
Uncaught ReferenceError: regeneratorRuntime is not defined
```
A quick search suggests it may be an issue with `async`,`await` per this SO answer:
https://stackoverflow.com/a/48534963/59913

( We just added those to platform in https://github.com/OriginProtocol/platform/pull/25 )

I blindly tried some of the answers in the SO question but to no avail. Lets talk through this tonight if we can.

wanderingstan:
@that1guy Doesn't this create a circular loop?
> run `npm link @originprotocol/origin` in platform.

I get this: 
```
$ npm link @originprotocol/origin
/Users/stan/Origin/platform/packages/origin.js/node_modules/@originprotocol/origin -> /usr/local/lib/node_modules/@originprotocol/origin -> /Users/stan/Origin/platform/packages/origin.js
```

I ended up with the same error in the end, regardless. :(

Per discord comments in #engineering by @aiham  , sounds like it may be us needlessly trying to support old browsers. 
> aiham - Today at 4:39 PM
> Yeah the SO answer is probably right.
> So what browser compatibility do we aim to achieve?
> If we only want to support evergreen browsers then we shouldn't need transform-async-to-generator
> I'll try to repro the error first.

that1guy:
Relevant to `Uncaught ReferenceError: regeneratorRuntime is not defined`: https://github.com/babel/babel/issues/5085

@wanderingstan my `npm link` steps were a bit off.  Apologies.

1) You're supposed to to `npm link` while inside the directory `platform/packages/origin.js`
2) Then you need to run `npm link @originprotocol/origin` while inside the root directory of `origin-demo-dapp`

Regardless, the `Uncaught ReferenceError: regeneratorRuntime is not defined` is unrealted to `npm link` I think.

that1guy:
@wanderingstan here is fix for `Uncaught ReferenceError: regeneratorRuntime is not defined`: https://github.com/OriginProtocol/platform/pull/31/commits/12552bc52190c5909f0b87c1929185e29867dbff

PR #42, opened 2018-03-16
Convert service methods to async/await

aiham:


wanderingstan:
Hey @aiham , working with @crazybuster yesterday we hit an issue where `submitListing()` in `origin-service.js` was returning a promise rather than the tx of a transaction. I think we were just missing the `await` here:
https://github.com/OriginProtocol/platform/blob/3acd2e93d5c8d467f3b8d928d900bda0b5b0736b/packages/origin.js/src/origin-service.js#L29

I put that in and things seem happy, but wanted to double check that you didn't have something else in mind at the time. (Times like this also make me wish js was a typed language. :)

aiham:
@wanderingstan Yes that's exactly what I should have done! Sorry about that.

aiham:
@wanderingstan Unfortunately I hadn't gotten around to writing proper tests in https://github.com/OriginProtocol/platform/blob/develop/packages/origin.js/test/origin-service.test.js so I didn't catch that one.

Issue #5, opened 2018-02-22
Implement user identity for Twitter publicly-auditable proof


matthewliu:

tyleryasaka:
As can be seen in the activity timeline here, I have created a PR that implements some methods for signing and verifying messages.

However, I've run into a roadblock on the Twitter implementation. Twitter's API requires authentication for all requests (even read-only requests for publicly available data). All of their supported authentication methods require some sort of API key (in addition to any client-side keys that are required for permissioned actions such as posting a new tweet), which means a server has to be involved. Currently our origin.js library is purely client-side.

What is our strategy for implementing publicly auditable proofs for Twitter? Should we add a separate open source server library? Or just set up our own server and provide it as a service? Or some other approach maybe?

I'll need some feedback on this before I can continue.

joshfraser:
Thought we might run into this problem. 3 options. None of them are great.

1) we set up a twitter-proxy that the dapp connects to that uses our keys. it's important that we are merely forward the contents of the tweet versus verifying it server-side to keep things as trustless as possible.

2) we just set up an account and publish our keys to the world. high risk that Twitter won't like it and will deactivate our keys especially if someone else runs across them and starts abusing them. quick and dirty hack to get it working, but probably not a viable long term solution.  https://stackoverflow.com/questions/17049821/setting-up-twitter-api-getting-the-last-few-tweets

3) we do a proper oauth handshake to connect each user and then use their personal oauth keys to do the fetching. one challenge would be that you would need to connect your own twitter account before you can fetch and verify other people's accounts.

I think our best option is probably #1. I hate adding centralized servers into the mix, so it's important that we keep it open-source & make it possible for people to swap our the verification server that they want to use. We should generalize this of course, as twitter won't be the only place where we run into this problem.

Anyone else want to weigh in?

joshfraser:
There's a few places where a centralized server is necessary:

- Indexing server (search)
- Notification server (sending notifications when you get a booking)
- Various proxies (fetching tweets, proxying Civic responses, etc)

Maybe we combine this all in one place and then you have 1 single setting to choosing your preferred centralized provider. Origin Protocol can be the default of course, but you should be able to swap it out for any other provider you want, including hosting it yourself. 

Need a good name for this thing. Ideas?

tyleryasaka:
Agreed on all points. Hmm, maybe we could call it something like `supporting server`? (Just throwing an idea out there.)

matthewliu:
[...]

PR #44, opened 2018-03-16
Publicly-auditable Twitter Proofs

tyleryasaka:
*WIP: need to wait until https://github.com/OriginProtocol/platform/pull/78 is merged and then do some refactoring*

This implements methods for using publicly-auditable twitter proofs. Users of the library now have a method to get a proof message to tweet, a method to create a proof, and a method to get a validated twitter username (if the user has any twitter proofs).

This requires the use of a centralized server (see https://github.com/OriginProtocol/platform/issues/5). I've implemented a possible first iteration of such a server: https://github.com/tyleryasaka/origin-server. I tested and the code in this PR works with the code in that server repo. For now the server url is just passed into each method call, but the next thing I would want to do is set up origin.js to take configuration options at initialization, and to be able to set the centralized server url at initialization time.

~Creating a twitter proof does require access tokens that belong to individual users. Currently these are passed in to the method call. In the demo-dapp, we'll have to use an interface to get these from the user (maybe using something like [auth0](https://auth0.com/)).~

Lastly I'll mention that I'm aware the origin.js service methods need some refactoring. There's some duplication, some inconsistency, methods that probably don't need to be exposed publicly, and overall a cluttered structure due to a bunch of us doing work on these services simultaneously. Once we have the server architecture in place I would definitely like to do some refactoring. At the moment I want to be cautious about making too many changes at once, especially changes that would break the current demo-dapp.

Refs: #5 

### Example usage:

Adding a proof:
```
twitterIdentityService.addProof('originprotocol', "http://localhost:4000")
.then(() => {
  console.log('added proof')
})
.catch((err) => {
  console.log('error', err)
})
```

Getting a verified username:
```
twitterIdentityService.getValidatedUsername(window.web3.eth.accounts[0], "http://localhost:4000")
.then((validatedUsername) => {
  console.log('validated username:', validatedUsername)
})
.catch((err) => {
  console.log('error', err)
})
```

### DIFF
--- a/packages/origin.js/package-lock.json
+++ b/packages/origin.js/package-lock.json
@@ -2714,6 +2714,20 @@
       "integrity": "sha1-Qa4u62XvpiJorr/qg6x9eSmbCIc=",
       "dev": true
     },
+    "ethereumjs-util": {
+      "version": "5.1.5",
+      "resolved": "https://registry.npmjs.org/ethereumjs-util/-/ethereumjs-util-5.1.5.tgz",
+      "integrity": "sha512-xPaSEATYJpMTCGowIt0oMZwFP4R1bxd6QsWgkcDvFL0JtXsr39p32WEcD14RscCjfP41YXZPCVWA4yAg0nrJmw==",
+      "requires": {
+        "bn.js": "4.11.7",
+        "create-hash": "1.1.3",
+        "ethjs-util": "0.1.4",
+        "keccak": "1.4.0",
+        "rlp": "2.0.0",
+        "safe-buffer": "5.1.1",
+        "secp256k1": "3.5.0"
+      }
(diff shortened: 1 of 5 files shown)


wanderingstan:
@joshfraser  @tyleryasaka -- This is some great work, but I suspect we shouldn't merge into `develop` until we are clearer on what we're doing with the bridge server (aka central server). 

I'm thinking we put this in a `topic/identity/twitter-proofs` branch or something similar and then adjust it from there into our final identity solution. 

Thoughts?

tyleryasaka:
It's still going to be some time before this is ready. I'm going to just merge this into the topic branch I created and then create a new PR from that branch into `develop` when it's actually ready for review.

Issue #50, opened 2018-03-20
Test creation of Purchase contract


wanderingstan:
We need to figure out how to capture solidity events in our tests. In first case, we need to get the address of the created `Purchase` contract. 

See notes and initial research links here:
https://github.com/OriginProtocol/platform/blob/develop/packages/contracts/test/TestListing.js#L45-L49

DanielVF:
@wanderingstan It seems like it might be important for Listings to keep track of their Purchases. 

Otherwise, anyone can create any old custom Purchase contract and claim it belongs to listing. This would be bad.

Secondly, if Listings don't know their Purchases, then it requires tracking the entire blockchain to know the status of a listing's purchases. A single link from a Purchase to a listing, like we have now, would be the way it's handled in a relational database, but the blockchain only really lets you follow links one way. If you want bi-directional linking, then you have to link bi-directionally.

So if we do track a Listing's Purchases, then we don't have to watch for events here in the test, since we can just ask the listing for the most recent Purchase.

DanielVF:
Here's the code to pull a contract from the emitted event:

    const buyTransaction = await instance.buyListing(
      unitsToBuy,
      { from: accounts[1], value: 6 }
    )
    const listingPurchasedEvent = buyTransaction.logs.find((e)=>e.event=="ListingPurchased")
    const purchaseContract = purchaseDefinition.at(listingPurchasedEvent.args._purchaseContract)

I've added this, and am now working on more purchase tests.

PR #52, opened 2018-03-22
Purchases were being created with the wrong buyer.

DanielVF:
Purchases were being created with the wrong buyer.

`msg.sender` is the address of the function *caller*, so in this case, the Purchase contract was recording the buyer as the Listing contract not the user, since the Listing contract was the one that called the constructor.

This also shows how you can fetch a contract from an event fire during a transaction.


DanielVF:
This closes #50

DanielVF:
I'm closing this PR, since this code has been rolled into the api_spike branch

Issue #37, opened 2018-03-14
Define script or steps for Testnet deployment


wanderingstan:

wanderingstan:
This seems to be the nice way: 
http://truffleframework.com/tutorials/using-infura-custom-provider

Just have to put the mnemonic and infura id's in env or config variables.

wanderingstan:
Fixed in https://github.com/OriginProtocol/platform/pull/63

PR #63, opened 2018-03-27
env variables for deploying to ropsten, rinkeby, or mainnet

wanderingstan:
Trying to make testnet deployment easy and script-able. 

For: https://github.com/OriginProtocol/platform/issues/37

- Are environment variables are the right place to store wallet mnemonics? 
- Should we have just one mnemonic, since account keys work on any network?

### DIFF
--- a/packages/contracts/package-lock.json
+++ b/packages/contracts/package-lock.json
@@ -3,10 +3,85 @@
   "requires": true,
   "lockfileVersion": 1,
   "dependencies": {
+    "abstract-leveldown": {
+      "version": "2.6.3",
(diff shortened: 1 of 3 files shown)


nick:
Looks good. I've actually found the INFURA_ACCESS_TOKEN to be optional... but probably best to include it anyway.

DanielVF:
Looks good to me. 

These env vars are only used for deploys, right? So we should be good from a security perspective.

I think multiple mnemonics is good.

wanderingstan:
Yes, just for deploys. 
Once we're live on mainnet, I'd feel a little uncomfortable having the main wallet mnemonic existing anywhere long-term in my computer (e.g. in env var) but we can deal with that later.

PR #64, opened 2018-03-28
Update develop from master

tyleryasaka:
Merges upstream changes in master into develop.

I believe https://github.com/OriginProtocol/platform/pull/18 was created when master was the default branch, but merged after develop became the default. Because of this develop is behind master (which should never be the case). This gets it caught up.

### DIFF
--- a/packages/contracts/contracts/UserRegistry.sol
+++ b/packages/contracts/contracts/UserRegistry.sol
@@ -9,36 +9,31 @@ contract UserRegistry {
     * Events
     */
 
-    event NewUser(uint _index);
+    event NewUser(address _address);
 
     /*
     * Storage
     */
 
-    // Array of all users
-    userStruct[] public users;
+    // Mapping of all users
+    mapping(address => userStruct) public users;
 
     /*
     * Structs
     */
 
     struct userStruct {
-        address owner;
         bytes32 ipfsHash;
+        bool isSet;
     }
 
     /*
     * Modifiers
     */
 
-    modifier isValidUserIndex(uint _index) {
-        require (_index < users.length);
+    modifier isValidUserAddress() {
+        require (users[msg.sender].isSet);
         _;
-    }
-
-    modifier isOwner(uint _index) {
-      require (msg.sender == users[_index].owner);
-      _;
     }
 
     /*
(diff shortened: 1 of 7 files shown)


aiham:
Thanks for this @tyleryasaka

wanderingstan:
Yikes, thanks for catching this. 

I see that `npm test` fail with this error:
```
ReferenceError: formListingJson is not defined
    at IpfsService.submitFile (bundle.js:139237:38)
```
Which is caused by us no longer having the json when caching what we send to ifps, on this line:
https://github.com/tyleryasaka/platform/blob/311b8df4cc7b83d6bb1f95cb2ab5d9510234b5e4/packages/origin.js/src/ipfs-service.js#L78

We could do something like
```
this.mapCache.set(ipfsHashStr, JSON.parse(file.content))
```
but it seems dumb to re-parse what we just stringify'd. 

The real solution is, I think, to make the IPFS service totally agnostic about the kinds of data we're storing or retrieving. 

Put my changes in this commit:
https://github.com/OriginProtocol/platform/commit/6c8135ba3408380c143642330d1f869055d0faa2

What do you think?

And @tyleryasaka , are you not able to make branches on this repo yet? That should be the preferred way so that I can propose changes to a PR.

aiham:
@wanderingstan 

I think when we do a merge the merge branch should only contain the source branch's commits plus any commits required to resolve conflicts. Any other refactor/design changes should be included in a separate PR, this really helps with tracking when changes happened and for what reason.

As for `submitFile`, I think it's a good idea to abstract that stuff out of the IPFS service so it doesn't care what kind of data is being retrieved/stored. Maybe `submitFile` can also accept a path as a parameter so its not always the same file path.

tyleryasaka:
@wanderingstan Oops sorry. I'll start using branches within this repo for future PRs.

Yeah I was just running into that same issue. I pushed your fix to this branch. (Really like that solution by the way.)

I also pushed 2 other commits to keep everything in a working state. I removed one method that got added that we're not using anywhere yet, and that I wasn't sure about adding in the first place. I also added a fix for the `setUser` method.

I have questions about how we want to structure the platform services going forward, so I'm not sure whether the origin service is the best place to put the `setUser` method. But I figure we can have a separate discussion to deal with that.

wanderingstan:
@aiham I'm curious on your take on git strategy: On one hand the PR wasn't passing tests, so it felt odd to accept it that way. But to fix it required a bit of refactor/design. (The last years I've been on small prototyping teams and when I had my company we were still in the dark ages of svn. Shudder!)

aiham:
@wanderingstan I'd personally make the minimal change to get it passing, because of course you don't want develop to be broken. But then right after get working on making the improvement/refactor in a new PR. This is just my preference to keep PRs distinct and focused. Thanks for asking!

wanderingstan:
Made the update to demo-dapp to reflect the changed method name:
https://github.com/OriginProtocol/demo-dapp/commit/785f26b32cec12839a772b6a52fdb850be869691
`develop` on both branches are working together as far as I can tell!

@aiham makes sense. I can see where a PR could get a domino effect of tons of changes, so makes sense to keep them razor focussed.

Issue #39, opened 2018-03-14
Convert UserRegistry to use map instead of array


wanderingstan:
With a mapping we can look up a user by indexing off their wallet address. 

In current form, you'd have to walk the array to find a particular user. 

@tyleryasaka : any reason why this would be a bad idea?

tyleryasaka:
@wanderingstan Sorry just saw this. Funny, when you posted this I had already created a PR which turns it into a map instead of an array 😄. You merged this recently so I think we can close this issue.

https://github.com/OriginProtocol/platform/blob/master/packages/contracts/contracts/UserRegistry.sol#L19

wanderingstan:
@tyleryasaka My turn to be sorry. 

One thought: Do we need the `isSet` bool and related struct?, [here](https://github.com/OriginProtocol/platform/blob/master/packages/contracts/contracts/UserRegistry.sol#L25-L28)? I'm thinking we could just have a straight mapping and (I think) save the 256 bits needed for the bool:
```
    // Mapping of all users
    mapping(address => bytes32) public users;
``` 
In the mapping, if there is nothing at the address, it will return 0x00 indicating that we have nothing on that wallet address.

tyleryasaka:
@wanderingstan You're right. [This answer](https://ethereum.stackexchange.com/questions/13021/how-can-you-figure-out-if-a-certain-key-exists-in-a-mapping-struct-defined-insi) is what made me think this was necessary, as I had run into this scenario on another contract before. Reading it now though, I see that just checking for the 0 value will be sufficient for us.

wanderingstan:
Yeah, we have the advantage that the ipfs hash will never be zero. (Although that's an interesting question: I suppose its theoretically possible (lie a 1-in-a-zillion chance) that a document would hash to 0x00000000000000000000000000000000, but that's a risk we can afford!

wanderingstan:
This is fixed as part of big pull https://github.com/OriginProtocol/platform/pull/44/

Will close when that gets merged.

tyleryasaka:
Closing as all the issues discussed here have been resolved.

PR #66, opened 2018-03-29
Refactor user registry to just be a mapping of addresses to ipfs hashes

tyleryasaka:
Simplifies the users mapping in the user registry contract, per the discussion in https://github.com/OriginProtocol/platform/issues/39.

### DIFF
--- a/packages/contracts/contracts/UserRegistry.sol
+++ b/packages/contracts/contracts/UserRegistry.sol
@@ -16,23 +16,14 @@ contract UserRegistry {
     */
 
     // Mapping of all users
-    mapping(address => userStruct) public users;
-
-    /*
-    * Structs
-    */
-
-    struct userStruct {
-        bytes32 ipfsHash;
-        bool isSet;
-    }
+    mapping(address => bytes32) public users;
 
     /*
     * Modifiers
     */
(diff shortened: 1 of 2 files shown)


packages/contracts/contracts/UserRegistry.sol
```diff
@@ -16,23 +16,14 @@ contract UserRegistry {
...
-        bytes32 ipfsHash;
-        bool isSet;
-    }
+    mapping(address => bytes32) public users;
 
     /*
     * Modifiers
     */
 
     modifier isValidUserAddress() {
```
wanderingstan: This modifier does not appear to be used anywhere, and the logic of it seems strange: In order to be valid, the `users` mapping for the user must be empty/zero?
If this is really needed, a comment here would be appropriate. 
https://github.com/OriginProtocol/platform/blob/36335ca374165524f18c0bb29acacf2d88e562af/packages/contracts/contracts/UserRegistry.sol#L25-L28

tyleryasaka: @wanderingstan You're right. That was leftover from an older version of this contract. Making a PR to remove it.

PR #68, opened 2018-03-31
Feature/2018 03 28 api spike

DanielVF:
New API


wanderingstan:
We've got ;'s in all the new `.js` files.  I'm not huge a fan of the `npm` style either, but its what we've got going so we should stay consistent.

DanielVF:
Yeah. After we do all the current platform merging, we can update the rest of the files to be consistent.

wanderingstan:
Thanks for leading the charge on our public-facing API!

PR #79, opened 2018-04-04
🚚Move Purchase contract creation out into a library

DanielVF:
This means that every Listing no longer has to contain the entire Purchase contract binary code. This makes a big cut in the cost to create a new listing.

This also reduces the cost of deploying the ListingRegistry from 4.7 million (on the edge of the limit) to 2.7 million.


DanielVF:
@tyleryasaka I moved up a previous commit and didn't alter that comment. This code is good now.

PR #85, opened 2018-04-09
✨Python ingest support for bridge server

DanielVF:
This code adds some events for the bridge server to listen to, and the start of python API for working with the origin blockchain data.

It is expected we'll remove the python code from this repo once it's up and going on the bridge server.


joshfraser:
Agreed, let's get this started on the bridge server. @tyleryasaka you probably have more context on how that is setup if you want to tackle porting it over.

tyleryasaka:
@Gzing This is the part of the bridge server you're working on I believe. I'll let you integrate this with the work you're doing as needed.

packages/originprotocol-python/originprotocol/origin.py
```diff
@@ -0,0 +1,66 @@
+import json
+import time
+from web3 import Web3
+from web3.contract import Contract
+
+
+class Origin:
+    def __init__(self, rpc_server, ipfs_server="https://gateway.originprotocol.com/"):
```
amites: `str` as default value is mutable, better to use `None` and set in method

DanielVF:
@amites Thanks for the suggested changes! This is useful.

It looks like Gagan has already pulled the Python API portion of this PR into the bridge_server repo, so I'm closing this PR, and moving just the needed contract events into another PR.

PR #86, opened 2018-04-10
✨Purchase API for origin.js

DanielVF:


DanielVF:
Thanks @tyleryasaka! I've added `purchases`  to the origin class.

PR #87, opened 2018-04-11
✨Shipping Pending stage, contract and API

DanielVF:
- Change origin.js to use strings for stages, rather than numbers
- Add new shipping_pending state to contracts, contract tests, origin.js, and origin.js tests
- Add sellerConfirmShip to contracts, contract tests, origin.js, and origin.js tests
- Rename getPayout to collectPayout, since it not a getting data method
- Add PurchaseChance events for Python ingest support

### DIFF
--- a/packages/origin.js/src/ResourceBase.js
+++ b/packages/origin.js/src/ResourceBase.js
@@ -3,13 +3,34 @@ class ResourceBase {
     this.contractService = contractService
     this.ipfsService = ipfsService
   }
-
-  async contractFn(address, functionName, args = [], value = 0) {
+  /**
+   * Runs a call or transaction on a this resource's smart contract.
+   *
+   * This handles getting the contract, using the correct account,
+   * and building our own response for origin transactions.
+   *
+   * If doing a blockchain call, this returns the data returned by
+   * the contract function.
+   *
+   * If running a transaction, this returns an object with a
+   *   - tx - transaction hash
+   *   - whenFinished - a promise that resolves when the transaction is mined
+   *
+   * @param {string} address - address of the contract
+   * @param {string} functionName - contract function to be run
+   * @param {*[]} args - args for the transaction or call.
+   * @param {{gas: number, value:(number | BigNumber)}} options - transaction options for w3
+   */
+  async contractFn(address, functionName, args = [], options = {}) {
+    // Setup options
+    const opts = Object.assign(options, {}) // clone options
+    opts.from = await this.contractService.currentAccount()
+    // Get contract and run trasaction
     const contractDefinition = this.contractDefinition
     const contract = await contractDefinition.at(address)
-    const account = await this.contractService.currentAccount()
-    args.push({ from: account, value: value })
+    args.push(opts)
     const transaction = await contract[functionName].apply(contract, args)
+    // Decorate transaction with whenFinished promise
     if (transaction.tx != undefined) {
       transaction.whenFinished = async () => {
         await this.contractService.waitTransactionFinished(transaction.tx)
--- a/packages/origin.js/src/ResourceBase.js
+++ b/packages/origin.js/src/ResourceBase.js
@@ -1,0 +1,22 @@
+class ResourceBase {
+  constructor({ contractService, ipfsService }) {
+    this.contractService = contractService
+    this.ipfsService = ipfsService
+  }
+
+  async contractFn(address, functionName, args = [], value = 0) {
+    const contractDefinition = this.contractDefinition
+    const contract = await contractDefinition.at(address)
+    const account = await this.contractService.currentAccount()
+    args.push({ from: account, value: value })
+    const result = await contract[functionName].apply(contract, args)
+    if (result.tx != undefined) {
+      result.whenMined = async () => {
+        await this.contractService.waitTransactionFinished(result.tx)
+      }
+    }
+    return result
+  }
+}
+
+export default ResourceBase
--- a/packages/origin.js/src/ResourceBase.js
+++ b/packages/origin.js/src/ResourceBase.js
@@ -9,13 +9,13 @@ class ResourceBase {
     const contract = await contractDefinition.at(address)
     const account = await this.contractService.currentAccount()
     args.push({ from: account, value: value })
-    const result = await contract[functionName].apply(contract, args)
-    if (result.tx != undefined) {
-      result.whenMined = async () => {
-        await this.contractService.waitTransactionFinished(result.tx)
+    const transaction = await contract[functionName].apply(contract, args)
+    if (transaction.tx != undefined) {
+      transaction.whenFinished = async () => {
+        await this.contractService.waitTransactionFinished(transaction.tx)
       }
     }
-    return result
+    return transaction
   }
 }
 
--- a/packages/origin.js/src/resources/listings.js
+++ b/packages/origin.js/src/resources/listings.js
@@ -1,10 +1,11 @@
 // For now, we are just wrapping the methods that are already in
 // contractService and ipfsService.
+import ResourceBase from"../ResourceBase"
 
-class Listings {
+class Listings extends ResourceBase{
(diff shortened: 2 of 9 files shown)


packages/origin.js/src/ResourceBase.js
```diff
@@ -0,0 +1,22 @@
+class ResourceBase {
```
tyleryasaka: nice

packages/origin.js/src/resources/purchases.js
```diff
@@ -1,48 +1,56 @@
...
-            amountWei
-        )
+    this._STAGES_TO_NUMBER = {
+      awaiting_payment: 0,
+      shipping_pending: 1,
+      buyer_pending: 2,
+      seller_pending: 3,
+      in_dispute: 4,
+      review_period: 5,
+      complete: 6
```
tyleryasaka: I don't think consumers of origin.js should have access to these stages as attributes. They should be able to see the string version of a particular stage value when interacting with origin.js, rather then the integer values, of course. But they should not have to access any properties on the purchase resource that just define the stages.

Because these are only for internal use, I'm pretty sure they should be defined as constants at the top of the file. Shouldn't be placed anywhere on the exported object.

(If we want to talk about sharing between other internal files, we can get fancier and define modules that simply contain constant definitions to share between files internally. But I don't think we need that yet.)

packages/origin.js/test/resource_purchases.test.js
```diff
@@ -56,45 +59,69 @@ describe("Purchase Resource", function() {
...
+  // -----
+
   describe("simple purchase flow", async () => {
     before(async () => {
       await resetListingAndPurchase()
     })
 
     it("should get a purchase", async () => {
-      expect(purchase.stage.toNumber()).to.equal(0)
+      expectStage(purchases.STAGES.AWAITING_PAYMENT)
```
tyleryasaka: Corresponding with my other comment, I don't think we should be accessing our constant definitions from the class that we're testing.

1. It's not good design to export something just so it can be used in a test
2. Our tests should not be retrieving values to expect from the very code they are testing. If we make a mistake in one of those definitions, the tests will not catch it because they are using the exact same definitions.

So although it is a bit redundant, I believe it's best to simply define the stages as constants at the top of our test file. (We're already doing this in the contract tests.) At the cost of having to do a bit more typing when changing values, it ensures that our tests will not be fooled by mistakes in our source. If someone updates a constant in a source file but forgets to update the tests, worst that happens is that tests fail. 😃

packages/origin.js/src/ResourceBase.js
```diff
@@ -0,0 +1,22 @@
...
+  }
+
+  async contractFn(address, functionName, args = [], value = 0) {
+    const contractDefinition = this.contractDefinition
+    const contract = await contractDefinition.at(address)
+    const account = await this.contractService.currentAccount()
+    args.push({ from: account, value: value })
+    const result = await contract[functionName].apply(contract, args)
+    if (result.tx != undefined) {
+      result.whenMined = async () => {
```
tyleryasaka: See my comment here: https://github.com/OriginProtocol/platform/issues/91

packages/origin.js/test/resource_purchases.test.js
DanielVF: I think I see what you are saying here - we want to make sure that what we "promise" in the API in terms of string stage names  is tested so that we don't accidentally change it, and break downstream code.

I've added a section to the tests that checks that these do not change.

The stage-name-strings-as a constant isn't just intended for our tests - it's for downstream code to use too. It's a pretty common javascript idiom for enum-like strings to be accessed through namespaced constants like this.

packages/origin.js/src/resources/purchases.js
```diff
@@ -1,48 +1,57 @@
...
+class Purchases extends ResourceBase{
+  constructor({ contractService, ipfsService }) {
+    super({ contractService, ipfsService })
+    
+    this.contractDefinition = this.contractService.purchaseContract
 
-    async buyerConfirmReceipt(address){
-        return await this.contractFn( address, 'buyerConfirmReceipt')
-    }
+    this.STAGES = {}
```
tyleryasaka: @DanielVF I still don't understand why we want to export the stage definitions as part of the API. I've never used an API that exported a static definition of something. Can you provide an example of why this needs to be part of the exported object?

DanielVF: 1) IDE's and TypeScript can autocomplete (if I actually explicitly listed the stages, rather than auto generating - which I may do later). This really helps those folks.

2) It gives library users a nice code-accessible list of all valid purchase stages.

3) It lets IDE's and other javascript tools to reason about all the places that are actually working with stages or to make automated refactoring.

4) Catches typos. Typos go straight to undefined.

5) Not applicable to this exact code, but in other situations, it allows a library author to swap out values without affecting downstream code.

tyleryasaka: 1. Autocomplete is nice but it's just an editor-specific convenience. I don't think APIs should cater to editors like that. This point seems like a nice side benefit rather not a reason in and of itself.

2. Same as the first point.

3. I totally agree about defining constants this way in order to catch typos. Definitely a good pattern to follow. However, I think it's the consumer's responsibility to do watch out for typos in its own codebase. The API should be agnostic about how its code will be used. (Providing it in the API implies that the client will be referencing our defined constants rather than just using literals - now we're making assumptions about how are API is used externally).

5. This smells like a bad pattern. Yes that is technically possible; however, a well-designed API should not be changing these definitions except during major releases. Changing definitions should be a breaking change, and exporting a list of constant values that can actually change will encourage people to change these definitions in minor versions, with no mention of a breaking change. Now if someone happens to screw up and reference a literal value rather than using the exported list to make the conversion, their code doesn't work.

I'd be much more willing to accept this pattern if I could see an example of a respected code base using this pattern. I haven't seen it yet and I'm concerned it's not a good design pattern.

I hope I not blowing a small thing out of proportion, but I think it's an important question in terms of architecture design.

@wanderingstan Any thoughts? (To summarize, I am concerned that it's bad practice to be exporting constants in our API, especially as attributes on an instantiated class.)

tyleryasaka: @DanielVF Those were vague because I haven't thought them through all the way. Wasn't a very good argument. I think to me it comes down to this being a different pattern than any APIs I've ever worked with. I want to stick with conventions.

For example, Stripe's API has similar attributes that can be one of multiple values. But they don't provide any sort of object or array or whatever that lets you fetch the list of possible values in the code. They simply list them out in the docs.

This is just one example, took less than a minute to find.

https://stripe.com/docs/api#payout_failures

How do you feel about following Stripe's API design as a guideline?

DanielVF: Yes, exactly. I'm planning on always giving users of the api stages in string form, and I'm planning on always accepting stages from a user in string form.

DanielVF: Since it might not have been clear - the API as it currently is only lets the API consumer see stages-as-strings. The consumer never knows about stages-as-numbers.

packages/origin.js/src/resources/listings.js
```diff
@@ -1,10 +1,11 @@
 // For now, we are just wrapping the methods that are already in
 // contractService and ipfsService.
+import ResourceBase from"../ResourceBase"
```
tyleryasaka: Missing space after `from`

packages/origin.js/src/resources/purchases.js
```diff
@@ -1,48 +1,57 @@
...
+    Object.entries(_STAGES_TO_NUMBER).map(([k, v]) => {
```
tyleryasaka: We should be using `forEach` here, not `map`. `forEach` just iterates, while `map` returns a new array. We're not doing anything with the return value of this method call so we shouldn't be using `map`.

PR #95, opened 2018-04-14
One line to rule them all

tyleryasaka:
### Checklist:

### Description:

This is a huge restructure, but I think this moves us in the right direction.

1. One-line setup! Just `npm setup` and `npm start` to get it up and running and linked locally. Bonus: `npm run test:contracts` to run contract tests.
2. `npm start` really does everything:
	- starts ganache
	- deploys contracts using migrations
	- builds JS
	- watches for changes in contracts; change triggers redeployment, which triggers JS rebuild
	- watches for changes in JS; change triggers rebuild
	- makes browser tests automatically available at `localhost:8081`
3. Abstracts Truffle behind our own commands. We're still using it as much as we were before, but now we're calling it programmatically from our scripts. If we ever decide to drop Truffle (for example to be able to use the latest web3), we can do so without having to change our development process.
4. Simpler codebase structure. We now have just *one* npm package, and it's in the root of the repo. Makes codebase easier to understand and work with imho.
5. Removes env var dependencies. origin.js will always be injected, so configurable options like IPFS gateway are passed into the constructor at initialization.
6. Updates readme. I've merged everything into a single readme. There's probably too much in it at the moment. I think the next step would be to move some of the more general Origin Protocol-level stuff to a separate repo (maybe documentation?). This readme should be specific to the origin.js library.

New structure looks like:

- demo dapp
	- readme (contains the full setup instructions; references platform readme)
- platform/origin.js (the repo is still named platform, but it exports a single origin.js package at the top level)
	- readme (contains setup instructions just for origin.js)
	- contracts (just a directory, *not* a package; contains all of the truffle code including the solidity contracts)

Also, see accompanying PR in demo dapp repo. The main change there is that it updates the readme to correspond to these changes. https://github.com/OriginProtocol/demo-dapp/pull/108

I've tested the process of walking through local setup using both readmes - but if other people could also try it out that would be great!


wanderingstan:
So cool!

Just a little clarification in the readme and we're all set.
- It wasn't clear to me that I could run the demo dapp at same time tests were being served.
- I had `webpack compiling` repeated many times as last output from `npm start`. I assume that's it recompiling after file changes?
- For a newbie, it might not be clear that after `npm start` you can launch demo dapp. 

I might take a stab at the README in the morning. 

But again: Awesome to have this! This makes the onboarding experience light-years beyond our old mess of switching tabs and zillion steps.

wanderingstan:
On [truffle.js](https://github.com/OriginProtocol/platform/pull/95/files#diff-39191e6f1314a8f369aa050fd5835a22R9) , can we actually change the port to 8545? That's the default port used by metamask, and would save people from having to configure a new RPC URL.

tyleryasaka:
@wanderingstan Yes we absolutely can! I thought about doing that here but decided to hold off as there is already a lot going on. I'll create a PR for that.

wanderingstan:
Yeah good call on making a separate PR for the port. I remember aiham 's good advice to keep PRs pretty focussed, and you this one (by necessity) was already doing a lot.

PR #113, opened 2018-04-20
✨Listing.get by address

DanielVF:
### Description:

Currently we only can get information about listing by the index number of where its address is stored in the ListingsRegistry contract. This is a left over of when when we used store all listing data in the ListingsRegistry contract.

We need to allow API users to directly get the Listing data from a listing contract address. This is a much more useful way of getting the listing, and will be the primary way listings are fetched in the future.

This adds an API method `origin.listings.get(listingAddress)`, and the need contract changes to use it.

It also adds a listing created date to the listing, since this is information we'll need to use from listings on the front end.

This should merged in after PR #111.

### Checklist:

Will do later tonight:

### DIFF
--- a/src/resources/listings.js
+++ b/src/resources/listings.js
@@ -12,13 +12,34 @@ class Listings extends ResourceBase{
     return await this.contractService.getAllListingIds()
   }
 
+  async get(address) {
+    const contractData = await this.contractFn(address, "data")
+    let ipfsHash = this.contractService.getIpfsHashFromBytes32(contractData[1])
+    const ipfsData = await this.ipfsService.getFile(ipfsHash)
+
+    let listing = {
+      address: address,
+      ipfsHash: ipfsHash,
+      sellerAddress: contractData[0],
+      priceWei: contractData[2],
+      price: this.contractService.web3.fromWei(contractData[2], "ether").toNumber(),
+      unitsAvailable: contractData[3],
+      expiration: contractData[4].toNumber(),
+
+      name: ipfsData.data.name,
+      category: ipfsData.data.category,
+      description: ipfsData.data.description,
+      location: ipfsData.data.location,
+      pictures: ipfsData.data.pictures
+    }
+
+    return listing
+  }
+
+  // This method is DEPRCIATED
   async getByIndex(listingIndex) {
-    const contractData = await this.contractService.getListing(
-      listingIndex
-    )
-    const ipfsData = await this.ipfsService.getFile(
-      contractData.ipfsHash
-    )
+    const contractData = await this.contractService.getListing(listingIndex)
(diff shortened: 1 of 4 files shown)


micahalcorn:
Does it make sense to also return addresses instead of ids [here](https://github.com/OriginProtocol/origin-js/blob/master/src/contract-service.js#L80)?

wanderingstan:
@micahalcorn that entire method is a stop-gap until we get the indexer (on bridge server) working. That said, now that Listings are contracts (instead of structs) it is possible I think to return all the addresses from a solidity function.

wanderingstan:
Getting pretty pedantic, but I'd argue in [this section](https://github.com/OriginProtocol/origin-js/blob/d241b6f1488d0af8ae68d802dcd607899ad00c83/src/resources/listings.js#L24-L25) that `price` should be the price in wei, as that is the true un-corruptible price, and we should have `priceEth` which just a convenience, and has all the problems of floating point numbers. I.e. the shorter name is the more fundamental, the longer name is more processed. 

Relevant lines:
```
      priceWei: contractData[2].toString(),
      price: this.contractService.web3.fromWei(contractData[2], "ether").toNumber(),
```

DanielVF:
I'm thinking that we may want to keep our long term field name as "priceWie", with the currency in the name. This should help avoid confusion as to if it's ETH or USD.

I agree that we do need to do away with the "price" field returning ETH. I was just keeping it in for now for backwards compatibility.

DanielVF:
@micahalcorn Yeah, that whole getAllListingIds will be toast once we have the bridge server working and connected through the API.

src/resources/listings.js
```diff
@@ -21,7 +21,7 @@ class Listings extends ResourceBase{
       address: address,
       ipfsHash: ipfsHash,
       sellerAddress: contractData[0],
-      priceWei: contractData[2],
+      priceWei: contractData[2].toString(),
```
tyleryasaka: @DanielVF I'm confused - why are we returning `priceWei` as a string but `price` as a number?.

src/resources/listings.js
```diff
@@ -12,13 +12,35 @@ class Listings extends ResourceBase{
...
+      category: ipfsData.data.category,
+      description: ipfsData.data.description,
+      location: ipfsData.data.location,
+      pictures: ipfsData.data.pictures
+    }
+
+    return listing
+  }
+
+  // This method is DEPRCIATED
```
micahalcorn: Should be "deprecated", and it might help to have an explanation for when to remove the method. 😁

src/resources/listings.js
tyleryasaka: 👍gotcha

PR #117, opened 2018-04-23
Update repo link

tyleryasaka:
### Checklist:

### Description:

- Update references to repo name
- Remove outdated README for tests
- Enable migration to create sample data on all networks (so that our test nets will receive sample data)

Fixes https://github.com/OriginProtocol/origin-js/issues/108, https://github.com/OriginProtocol/origin-js/issues/109


contracts/migrations/3_create_sample_listings.js
```diff
@@ -3,13 +3,8 @@ var ListingContract = require("../build/contracts/Listing.json")
...
 
-module.exports = function(deployer, network) {
-  console.log("network", network)
-  if (network === "development") {
-    return deploy_sample_contracts()
-  } else {
-    // Don't create seed data on main net
-  }
+module.exports = function(deployer/*, network*/) {
+  return deploy_sample_contracts()
```
DanielVF: Sample contracts now include sample purchases, that are paid for with eth. This is just fine locally.

Will this change here affect  deployments? You wouldn't want to spend a surprise $2,000 USD when you deployed contracts to production.

tyleryasaka: @DanielVF Good point. These migrations are intended to never be run on the main net of course, but even on a test net that is a lot of Eth - and if we ever accidentally ran these on the main net it would be pretty disastrous.

I updated to only create the sample purchases locally. Sample listings will still be created on test nets.

PR #121, opened 2018-04-24
Improvements to README

jessieay:
### Description

* Start with step for Forking the repo, since that will be required for
all contributors without write access to the repo.

* Add correct URL for git cloning instructions (repo was renamed but
that name in the README did not change)

* Suggest cloning with SSH instead of HTTPS since that is how most
people access GitHub repos, in my experience (could be wrong on this
one).

* Use `1` for all items in ordered list so that they do not need to be
renamed if items are added. GitHub markdown will automatically increment
numbers in ordered lists.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,22 +77,24 @@ And buyers to:
 
  ### Local
 
+1. Fork the repo on GitHub
+
 1. Clone
-```
-git clone https://github.com/OriginProtocol/platform origin-platform && cd origin-platform
-```
+    ```
+    git clone git@github.com:YOUR_GITHUB_USERNAME/origin-js.git && cd origin-js
+    ```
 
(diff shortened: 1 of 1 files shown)


jessieay:
closing bc I just realized that I should PR against `develop` instead of `master` - can someone with admin privs make that the default branch so the PR behavior is more obvious?

tyleryasaka:
@jessieay We've had a back-and-forth discussion on what the default branch should be. I agree 100% that it's confusing to ask people to PR to a branch that's not default. The reasoning behind making the default `master` is that `develop` is less stable and more likely to be broken at various times (and we want things to work when newcomers try to clone our repo). It's not set in stone though; the confusion this caused for you is a good argument for making `develop` the default branch, imo.

tyleryasaka:
@jessieay Good point. I agree with you. I opened an issue so we can discuss further. https://github.com/OriginProtocol/origin-js/issues/123

PR #122, opened 2018-04-24
Improvements to README

jessieay:
### Description:

* Start with step for Forking the repo, since that will be required for
all contributors without write access to the repo.

* Add correct URL for git cloning instructions (repo was renamed but
that name in the README did not change). This is duplicative of
https://github.com/OriginProtocol/origin-js/pull/117 but I am including
it because of the following bullet point.

* Suggest cloning with SSH instead of HTTPS since that is how most
people access GitHub repos, in my experience (could be wrong on this
one).

* Use `1` for all items in ordered list so that they do not need to be
renamed if items are added. GitHub markdown will automatically increment
numbers in ordered lists.

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,22 +78,24 @@ And buyers to:
 
  ### Local
 
+1. Fork the repo on GitHub
+
 1. Clone
-```
-git clone https://github.com/OriginProtocol/platform origin-platform && cd origin-platform
-```
+    ```
+    git clone git@github.com:YOUR_GITHUB_USERNAME/origin-js.git && cd origin-js
+    ```
 
-2. Install:dev (shortcut for `npm install && npm link`). Linking makes this available as a local npm package for local dapp development.
- ```
- npm run install:dev
- ```
+1. Install:dev (shortcut for `npm install && npm link`). Linking makes this available as a local npm package for local dapp development.
+    ```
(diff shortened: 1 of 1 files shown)


jessieay:
Darn I was hoping that submitting a PR against `develop` would fix the Travis failure but it's still red. 

Given that my changes are only in the README, my guess is that the failure is not related to my changes and that any branch would fail the Travis webhook right now. Is that right?

README.md
```diff
@@ -78,22 +78,24 @@ And buyers to:
...
  ### Local
 
+1. Fork the repo on GitHub
+
 1. Clone
-```
-git clone https://github.com/OriginProtocol/platform origin-platform && cd origin-platform
-```
+    ```
+    git clone git@github.com:YOUR_GITHUB_USERNAME/origin-js.git && cd origin-js
```
tyleryasaka: Yeah, good point about `ssh` vs `https`. Now that you point it out, I think it's best not to assume what method the user will use to clone. I'm thinking we should just change this to say: 

`git clone` this repo

Or something like that.

Also, thanks for updating the reference to the repo!

joshfraser:
Thanks Jessie! Sorry about the failing tests. Not your fault!

Issue #110, opened 2018-04-20
Ungraceful failure when web3 is not present


micahalcorn:
This package [assumes that `web3` is available](https://github.com/OriginProtocol/origin-js/blob/61ce31e18069eb24af14969788e60bbe998b5920/src/contract-service.js#L21). Since there is no check, an exception of `Uncaught TypeError: Cannot read property 'currentProvider' of undefined` is thrown if it is not present.

In the DApp, we check for the presence of `web3` as a part of the rendering process but not before this package is imported.

**Is it the DApp's sole responsibility to provide `web3` prior to engaging with `origin-js` and/or should the package not assume that a global variable is present on the `window`?** I don't know whether or not it would appropriate for a package to yield anything other than a functional module or an exception.

cc @DanielVF @tyleryasaka

ambertch:
If it's the former, one option might be using this render callback (https://medium.com/@adamrackis/simplifying-life-with-react-render-callbacks-cb37d58e55) with where origin-js is loaded (https://github.com/OriginProtocol/demo-dapp/blob/master/src/components/app.js#L75)

ambertch:
With the latter, I think it's not unreasonable to have dependencies that are not immediately available during module import (a DB connection, for example).

I did a bit of research and saw an example of a dependency being injected into a component via triggering a re-render (by setting this.state).

DanielVF:
@micahalcorn 

Snipping some lines here:
```
constructor({ web3 } = {}) {
     this.web3 = web3 || window.web3
```
..snip..
```
     this[name].setProvider(this.web3.currentProvider)
```

So origin.js will first use a web3 that you pass into it, but if you don't pass in a web3, then it will check the window for one and use that. This seems pretty good to me, though we should probably show some kind of error in the console if neither are true.

A DAPP should be responsible for making sure there is a web3, and for showing a whatever UI error we need to show if there isn't a web3 available.

micahalcorn:
@nick we discussed this a little on Friday. I suppose that this might get resolved on the DApp side once we are supplying a version of `web3` (as opposed to relying on MetaMask's). As for the `origin-js` side, I think @tyleryasaka was going to take a look at this after knocking out some higher-priority issues.

tyleryasaka:
> Is it the DApp's sole responsibility to provide web3 prior to engaging with origin-js and/or should the package not assume that a global variable is present on the window?

[...]

PR #138, opened 2018-04-30
🔧 Use our own web3 instance

DanielVF:
Currently, origin.js uses whatever web3 is passed in to it. This is bad because different versions of metamask provide different versions of web3 with different APIs, thus breaking our code.

The solution is to use our own bundled version of web3, and route its transactions out through metamask’s web3, or whatever external web3 is provided to origin.js. This way we have a stable internal API. The [metamask FAQ](https://github.com/MetaMask/faq/blob/master/DEVELOPERS.md) recommends doing this: “You can use the injected Web3.js directly, but the best practice is to explicitly bundle the version of web3.js you used during development.”

### Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2767,6 +2767,22 @@
         "sha.js": "2.4.8"
       }
     },
+    "cross-fetch": {
+      "version": "2.1.1",
+      "resolved": "https://registry.npmjs.org/cross-fetch/-/cross-fetch-2.1.1.tgz",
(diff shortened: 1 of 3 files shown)


tyleryasaka:
Hey @DanielVF just FYI even though we're allowing web3 to be passed into the contract service, the origin-js package itself does not currently accept web3 as an option.

I did add this in an unmerged commit: https://github.com/OriginProtocol/origin-js/pull/128/commits/9973f063702b120178a7ec72bb3a8e517735b38e

DanielVF:
Yeah, @tyleryasaka - I figure we'll make small incremental changes here. First do this, then we can change how web3 is passed into origin, then we can change to using web3 1.0

micahalcorn:
Leaving a reference here to #110

PR #154, opened 2018-05-05
Migration fix

tyleryasaka:
We were getting the error mentioned in [this thread](https://ethereum.stackexchange.com/a/35780)

### Checklist:

This fixes the error @wanderingstan was getting while attempting to deploy to test nets.

We actually had this bug in the last deploy as well - only the migration that contained this code never ran, because we disabled it on everything except development net.

Used [this answer](https://ethereum.stackexchange.com/a/35780) to find the solution.

Original error:

```
$ truffle migrate --network rinkeby
Using network 'rinkeby'.

Running migration: 4_add_sample_issuers.js
  Running step...
Error encountered, bailing. Network state unknown. Review successful transactions manually.
Web3ProviderEngine does not support synchronous requests.
```

### DIFF
--- a/contracts/migrations/3_create_sample_listings.js
+++ b/contracts/migrations/3_create_sample_listings.js
@@ -9,12 +9,19 @@ module.exports = function(deployer, network) {
 }
 
 async function deploy_sample_contracts(network) {
-  console.log("deploying listings")
+  let accounts = await new Promise((resolve, reject) => {
+    web3.eth.getAccounts((error, result) => {
+      if (error) {
+        reject(err)
+      }
+      resolve(result)
+    })
+  })
 
-  const default_account = web3.eth.accounts[0]
-  const a_seller_account = web3.eth.accounts[1]
-  const a_buyer_account = web3.eth.accounts[2]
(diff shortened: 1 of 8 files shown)


contracts/migrations/3_create_sample_listings.js
```diff
@@ -9,12 +9,19 @@ module.exports = function(deployer, network) {
 }
 
 async function deploy_sample_contracts(network) {
-  console.log("deploying listings")
+  let accounts = await new Promise((resolve, reject) => {
```
nick: can also just say `let accounts = await web3.eth.getAccounts()`

tyleryasaka: Just tested, that doesn't work. I think it's because we are using Truffle's web3 here, which is version 0.x.

```
Error encountered, bailing. Network state unknown. Review successful transactions manually.
TypeError: Cannot read property '0' of undefined
    at deploy_sample_contracts (/Users/tyler/origin/origin-js/contracts/migrations/3_create_sample_listings.js:14:35)
    at <anonymous>
Using network 'development'.
```

I'm just going to leave it the way it is so that we can do this release.

PR #157, opened 2018-05-06
Add param for number of addresses

### LINKED ISSUE #156: Cannot deploy test listings to Ropsten/Rinkeby
When deploying to testnets, `3_create_sample_listings.js` fails with `Error: invalid address`. 

Investigation showed that problem was accounts beyond the first one were empty strings. In these lines:
https://github.com/OriginProtocol/origin-js/blob/develop/contracts/migrations/3_create_sample_listings.js#L21-L24

Problem is our `HDWalletProvider` in this line:
https://github.com/OriginProtocol/origin-js/blob/develop/contracts/truffle.js#L24-L27

Turns out it only manages one account. This appears to be a known limit [described here](https://github.com/trufflesuite/truffle-hdwallet-provider). 

> Currently, the HDWalletProvider manages only one address at a time, but it can be easily upgraded to manage (i.e., "unlock") multiple addresses.

I couldn't find how to unlock multiple addresses as they described, in short time I had this morning. Help appreciated.

micahalcorn:
Resolves #156.

### DIFF
--- a/contracts/truffle.js
+++ b/contracts/truffle.js
@@ -23,7 +23,7 @@ if (process.env.MAINNET_MNEMONIC) {
     provider: function() {
       return new HDWalletProvider(
         process.env.MAINNET_MNEMONIC,
-        `https://mainnet.infura.io/${process.env.INFURA_ACCESS_TOKEN}`)
+        `https://mainnet.infura.io/${process.env.INFURA_ACCESS_TOKEN}`, 0, 4)
     },
     network_id: 1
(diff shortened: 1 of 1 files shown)


wanderingstan:
Ah, it won't let me merge your PR in with my change. So just closing. :(

micahalcorn:
@wanderingstan no problem, lesson learn. And please don't look at the demo-dapp code. 😬

four

PR #171, opened 2018-05-10
Readme File Updates / Support AVN/NVM

gaboesquivel:
- add a .node-version file for automatic node Version switching with avn/nvm.
- add a table of contents to readme file using doctoc.
- update and rearrange readme file contents for a better organization.


tyleryasaka:
This looks good to me, although I feel like someone else should look at the documentation changes before approving. Can you take a look @micahalcorn ?

.node-version
```diff
@@ -0,0 +1 @@
+v9.11.1
```
nick: Is this something that can go in package.json?

micahalcorn: And is `9.11.1` the appropriate number? I think I'm still running `8.5.0` just fine.

micahalcorn:
I like the TOC but I think @wanderingstan needs to be the one to review the copy edits.

.node-version
tyleryasaka: Yeah, and right now we're not even instructing users to use nvm. I'll mention that there is an [engines](https://docs.npmjs.com/files/package.json#engines) section in package.json, but this is not enforced unless the user has set their own `engine-strict` flag, which is `false` by default.

micahalcorn: @tyleryasaka I actually use [n](https://github.com/tj/n) when I'm not using [rvm](https://rvm.io/). 😂

gaboesquivel: Hey @nick, there's only one project that does automatic node version switching based on engines from package.json [Kikobeats/nodengine](https://github.com/Kikobeats/nodengine) , but it doesn't really work well.  `avn` can be used with both `nvm` and `n`. AVN works with .node-version [wbyoung/avn](https://github.com/wbyoung/avn)

nick: Shall we just remove this file for now so we can get the rest merged? My personal opinion is that it's not necessary - we can either say in the README which node version you should have or use the `engines` section in package.json as Tyler suggested (or both)

gaboesquivel: sure, I will remove it. I do find it very useful though. If you are working on different projects and contributing to different open source packages, automatic node version switching is great, that way you just cd the project dir and boom¡ you get the right the right version, no need to read each project's readme file and manually change it.

gaboesquivel: removed @nick

gaboesquivel: Just FYI, they are still working reading from `engines.node`  of the package.json https://github.com/creationix/nvm/pull/1535. N efforts to support this feature are staled https://github.com/tj/n/issues/192. That's why the .node-version files is still required. weird right, not sure why it's taking so long. ¯\_(ツ)_/¯

nick: Thanks. You can always use `.node-version` in your own environment by adding to `.git/info/exclude`

gaboesquivel: yeah. just found a package that's handy too.

gaboesquivel:
```➜  origin-js git:(develop) check-engine
Checking versions...

✘ node version is incorrect! Expected 9.11.1 but was v8.9.3.

 Environment is invalid!

Comment (gaboesquivel) [reply .node-version]: yeah. just found another package that's handy too.  
https://www.npmjs.com/package/check-engine
```
➜  origin-js git:(develop) check-engine
Checking versions...

✘ node version is incorrect! Expected 9.11.1 but was v8.9.3.

 Environment is invalid!
```

gaboesquivel:
This is not longer mergeable due to conflicts with https://github.com/OriginProtocol/origin-js/commit/6537a9766311a402797b34bc9427a2ece01fdef1

I can incorporate those changes to this PR if necessary. 
It seems we want to:
- simplify the contributing section by removing invitation to the call and pointers to the dev notes
- simplify the local development section by refering to DApp readme and docs site instead.
- update wording of the introduction section and other couple lines

PR #181, opened 2018-05-12
Minify contracts

nick:
### Checklist:

### Description:

The contract JSON outputted by truffle contains lots of extra information such as AST, source, sourceMap, etc. It also includes information about the environment used to build the contracts including usernames and paths:

screen shot 2018-05-12 at 12 40 11 pm

Removing all this extra information results in a final build size, after minification, of 1.04Mb compared to the original size of 1.44Mb.

### DIFF
--- a/scripts/test.js
+++ b/scripts/test.js
@@ -7,12 +7,12 @@ const testFormat = require('./helpers/test-format')
 
 const start = async () => {
   try {
+    await testFormat()
     await startGanache()
     await testContracts()
     await deployContracts()
     await startIpfs()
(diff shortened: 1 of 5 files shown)


scripts/test.js
```diff
@@ -7,12 +7,12 @@ const testFormat = require('./helpers/test-format')
 
 const start = async () => {
   try {
+    await testFormat()
```
nick: I moved this to the top as it sucks to wait minutes for the other tests to pass only to find you have a formatting error

Issue #149, opened 2018-05-02
Enable buyers and sellers to leave reviews


micahalcorn:
**Likely Requirements**
- [x] five star rating score
- [x] optional comment string
- [x] scoped to transaction rather than ~counterparty or~ product
- [x] timestamped
- [x] affecting transaction state logs

**Possible Considerations**

1. review period (and expiration)
2. review reciprocity
3. preventing funds withdrawal

Contract changes:
- create review with buyerConfirmReceipt
- create review with sellerCollectPayout

Reviews API:
- reviews.find({purchaseAddress?:, userAddress?:, sellerAddress?:, buyerAddress?: listingAddress?:})

micahalcorn:
On Airbnb, I think reviews only ever become visible if both parties participate. Will we have a similar model? If so, how do we hang on to a review without publishing it to the blockchain until the counterparty has reciprocated?

micahalcorn:
I believe that the plan was for a purchase to not enter `complete` status until (a) the buyer has left a review, or (b) the review period has expired. Is that the case? If a dispute is created, can a review still be left or does it send the purchase state down a different path? When a buyer confirms that they have received a shipment, that should not automatically allow the seller to withdraw funds.

DanielVF:
Yes, reviews will definitely need to be indexed. 

We'll need the indexing server to check the contract to determine if the review is "valid", since we need to send the review to the blockchain when they happen, even if the other party still might open a dispute.

> "When a buyer confirms that they have received a shipment, that should not automatically allow the seller to withdraw funds." 

At that point the seller could either withdraw and review or open a dispute. They wouldn't be able to do both.

joshfraser:
Great question, @micahalcorn & there's no easy answer to this problem today. Storing / revealing private data on a public blockchain doesn't really work. NuCypher promises to help with this, but it's still under active development.

One temporary and unsatisfying solution would be to have the blind review process enforced entirely by origin.js so the reviews are hidden in the DApp till both sides have left a review. Of course, the data will already be public on IPFS & Ethereum, but you'll need to have some development chops to access it.

DanielVF:
@joshfraser Do we need reviews scores on-chain for any reason? If not, the review data we need to store would work as pure logs.

I'm thinking we don't need them on-chain. My guess is that any reputation system that operates deterministically on the blockchain could be gamed, and would be much slower than something off chain.

joshfraser:
[...]

PR #183, opened 2018-05-14
✨Purchase Reviews API

DanielVF:
Leaves ratings and optional textual reviews on purchases.

Every successful buyer and seller transaction ends up with a 1-5 rating, defaulting to 5.

Currently provides an API for getting the reviews from a single purchase. Will need bridge server integration to provide the rest of the search options.

#149 

### Checklist:

- [-] Update any relevant READMEs and [docs](https://github.com/OriginProtocol/docs)


src/resources/purchases.js
```diff
@@ -44,12 +46,34 @@ class Purchases extends ResourceBase {
...
+
+  async sellerGetPayout(address, data={}) {
+    const review = await this._buildReview(data)
+    const args = [review.rating, review.ipfsHashBytes]
+    return await this.contractFn(address, "sellerCollectPayout", args, {gas: 100000})
   }
 
-  async sellerGetPayout(address) {
-    return await this.contractFn(address, "sellerCollectPayout")
+  async _buildReview(data={}){
```
tyleryasaka: @DanielVF Maybe we should consider using a json schema here instead of doing our own validation?

We're already doing this for user profile: https://github.com/OriginProtocol/origin-js/blob/master/src/resources/users.js#L19

PR #188, opened 2018-05-15
Include identity address in user object

micahalcorn:
This exposes the identity address of a user in the [user object](http://docs.originprotocol.com/#user-get). This helps to properly satisfy https://github.com/OriginProtocol/demo-dapp/issues/158.

### Checklist:

### DIFF
--- a/src/resources/users.js
+++ b/src/resources/users.js
@@ -26,9 +26,10 @@ let validateUser = (data) => {
 }
 
 class UserObject {
-  constructor({ profile, attestations } = {}) {
+  constructor({ profile, attestations, identityAddress } = {}) {
     this.profile = profile
     this.attestations = attestations
+    this.identityAddress = identityAddress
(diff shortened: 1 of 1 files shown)


src/resources/users.js
```diff
@@ -51,6 +52,7 @@ class Users extends ResourceBase {
     let identityAddress = await this.identityAddress(address)
     if (identityAddress) {
       let userData = await this.getClaims(identityAddress)
+      userData.identityAddress = identityAddress
       return new UserObject(userData)
```
micahalcorn: For some reason, it didn't like the spread operator here. 😕

tyleryasaka: @micahalcorn Yep, we don't get to use the spread operator ☹️. I don't remember specifically why off the top of my head, but we can probably enable it with Babel if we want to sometime.

Issue #160, opened 2018-05-07
Create initial script for public net deployment


wanderingstan:
Eventually we want to automate as much of [the deployment checklist](https://github.com/OriginProtocol/origin-js/issues/104) as possible. 

A good first step would be:
- Check balance of our test accounts
- (If balance is too low)  
  - Get [test ethers for ropsten](https://github.com/sponnet/locals-faucetserver#api). When debugging a deployment issue you burn through test ethers pretty fast!
  - Not sure if Rinkeby has similar api. Worst case open this url: https://faucet.rinkeby.io/

Idea would be to use the env variables for wallet mnemonics (e.g. `RINKEBY_MNEMONIC`) along with `HDWalletProvider` (as in `truffle.js`) to get the wallet addresses. We need money in at least the first 2 accounts generated from the mnemonics. Ideally 4. 

Talk to me (@wanderingstan) if you want to tackle this, as I'm currently doing our deployments. I'll eventually get around to writing this myself if no one else jumps in.

digidigo:
@wanderingstan  - I can help with this one. Let's sync up so I can get some of your initial thoughts. I have confirmed that it is straightforward to grab eth balance from a script.

PR #189, opened 2018-05-15
Issue/160 initial deploy script

digidigo:
### Checklist:

### Description:

This PR attempts to lay some initial ground work for automating the deploy checklist. 
It should close issue #160 

First step is to use an environment variable to load an HDWallet for both Rinkby and Ropsten
Then check the balance of each address 
If the balance is lower that hardcoded limit of 3 eth then either 
Hit the Ropsten API or log a message to go to the Rinkby faucet


wanderingstan:
Thanks @digidigo !

Did a quick code-style cleanup to better match rest of repo. Might need a bit more too.

I'm thinking we should use same env vars as in [`truffle.js`](https://github.com/OriginProtocol/origin-js/blob/master/contracts/truffle.js), ie `RINKEBY_MNEMONIC` and `ROPSTEN_MNEMONIC`. 

A case could be made for having one unified mnemonic/wallet that is used across all networks, but I can't help thinking that our mainnet wallet will be *super* protected and not at all the same one we'd use on test nets.

scripts/deploy_checklist.js
```diff
@@ -0,0 +1,68 @@
...
+    }
+  }
+  return balances
+}
+
+const run = async () => {
+  var accounts = await getBalances(allAccounts)
+  for (var i = 0; i < accounts.length; i++) {
+    var account = accounts[i]
+    if ( account.eth < minimumEthForDeploy) {
```
gaboesquivel: By following No-Else-Return convention this code could be more readable and shorter ( fewer characters ). Rel https://github.com/OriginProtocol/origin-js/issues/193

tyleryasaka:
Closing as this appears to be stale. Feel free to reopen @wanderingstan

PR #190, opened 2018-05-16
Identity libraries for gas cost optimization

tyleryasaka:

### Checklist:

### Description:

Refactors solidity identity contracts to move as much logic as possible into libraries to reduce gas costs for deploying a new identity. I'm new to libraries so it's possible there are further improvements that can be made, but I think this is a good first pass.

Also uses a library for reporting gas estimates, swapping it for our in-house solution.

Reduced from ~2.7m to ~2.1m.

#### Before

screen shot 2018-05-14 at 2 49 01 pm

#### After

screen shot 2018-05-15 at 10 53 36 pm


contracts/migrations/2_deploy_contracts.js
```diff
@@ -1,8 +1,14 @@
-var ListingsRegistry = artifacts.require("./ListingsRegistry.sol");
-var Listing = artifacts.require("./Listing.sol");
-var UserRegistry = artifacts.require("./UserRegistry.sol");
-var PurchaseLibrary = artifacts.require("./PurchaseLibrary.sol");
-var OriginIdentity = artifacts.require("./OriginIdentity.sol");
+var ClaimHolder = artifacts.require("./ClaimHolder.sol")
```
gaboesquivel: use `const` instead of `var`. Related to https://github.com/OriginProtocol/origin-js/issues/192

src/services/contract-service.js
```diff
@@ -126,12 +131,26 @@ class ContractService {
     return new this.web3.eth.Contract(contract.abi, addrs)
   }
 
+  async getBytecode(contract) {
+    let net = await this.web3.eth.net.getId()
```
gaboesquivel: use `const`, relate to https://github.com/OriginProtocol/origin-js/issues/192

Issue #159, opened 2018-05-07
🎓What is a trusted listing contract?


DanielVF:
We need to decide on how we decided that Listing contracts are trusted for use in the Origin. An important part of this discussion is if there is a single source of truth for this, or can different DApps support a different set of Listing contracts?

### Why do we need to know what is a trusted contract

Anyone can create a contract that matches the interface of an Origin Listing contract, and can emit events that cause it to be picked up by Origin bridge servers. However, a contract that exactly matches the external interface, can still contain evil logic inside. For example, an otherwise normal looking listing contract could immediately allow the buyers money to be withdrawn and then close the listing, preventing disputes.

Similarly, any contract in the world could call a record transaction on a user review registry contract to record a review, even if it was just a custom contract that spammed out a five hundred five star reviews. A review registry has to be able to know if the contract calling it is trusted.

To avoid this, DApps, contracts, and the bridge server have to have a way to whitelist trusted listing contracts. This way when they see an event come in, or go to a listing, they can verify that the code behind that contract is trusted.

### Centralized or per DApp trust?

One possible approach is that only Origin gets to pick trusted contracts. Every DApp uses the same centralized contract to validate contracts. 

- Easy to ensure that all contracts use Origin token or lose the benefits of the data ecosystem.
- One central place that knows about all possible Origin contracts
- We probably have to write much bigger, configurable contracts.
- Probably encourages forking everything 

The alternative approach is to have each DApp maintain its own whitelist of trusted contracts bytecodes.

- Easier to allow innovation and very different DApps
- Doesn't provide a single point to remove all contracts of a given type from operation

### Implementation Options

This trusted list can be done with either a registry of trusted contract addresses or with a list of trusted contract bytecodes.

Thoughts, @joshfraser?

----

Immediately effects:

- https://github.com/OriginProtocol/origin-js/issues/149
- https://github.com/OriginProtocol/bridge-server/issues/44

joshfraser:
Thanks for opening this issue and putting this up for discussion. I don't know the right answer as there are pros and cons to both approaches as you laid out.

One thing I'll add is the possibility of us using a staking mechanism as a way of ensuring good behavior. For example, we could keep a very simple registry which keeps track of all of the listings on the Origin network, whether they're using our exact bytecode or not. The important thing is that you have to submit a large deposit that will be forfeited if you are proven to be a bad actor in some way. 

[...]

PR #201, opened 2018-05-21
✨Listing registry storage

DanielVF:
Per the discussion in #159 we're creating a way to store listings while allowing the creation and management code to be upgraded. We're also providing a constant time way to verify that a listing is trusted by a registry.

In the past, we've essentially thrown away all previous listings every time we've pushed a new version of the listing registry.

image

This splits off the listing storage, so that as we upgrade to new ListingsRegistry contracts, we can keep our list of past listing contracts. Other than the changes to the migration process, this is a drop in replacement - the listing registry still supports all the previous methods that it did before.

image

### New migration

Core listing address storage is now in a separate long term storage contract that can be reused by future ListingsRegistry contracts, and still have all the old contracts in use.

    // `storage` is the long lived storage contract
    const newRegistry = await deployer.deploy(ListingsRegistry, storage.address)
    await storage.setActiveRegistry(newRegistry.address)
   // And now we'll be using the new ListingsRegistry, with it's new contracts

### isTrustedListing()

You can check with the registry if a contract belongs to it:

    await listingsRegistry.isTrustedListing(listing.address)

### For later

- [ ] Origin.js checking isTrustedListing (not needed for current dapp features)
- [ ] Contract version numbers would be a separate PR.

### Checklist:


DanielVF:
Looks great to me.

PR #212, opened 2018-05-26
Tyleryasaka/listing address support

tyleryasaka:

### Checklist:

### Description:

1. Updates our `NewListing` event to include the new listing address.
2. Updates our `listings` resource to have an `allAddresses` method. Same as `allIds`, but with address instead of id

In one day, I ran into 2 separate issues that can be solved with these changes:
- Getting a listing address immediately after creating it. This is needed for clean and easy-to-use jsfiddle demos.
- Getting a list of all our listing addresses. This will allow us to hide listings by address seamlessly. See: https://github.com/OriginProtocol/demo-dapp/pull/222 and https://github.com/OriginProtocol/demo-dapp/issues/221

### DIFF
--- a/contracts/contracts/ListingsRegistry.sol
+++ b/contracts/contracts/ListingsRegistry.sol
@@ -12,7 +12,7 @@ contract ListingsRegistry {
    * Events
    */
 
-  event NewListing(uint _index);
+  event NewListing(uint _index, address _address);
 
   /*
    * Storage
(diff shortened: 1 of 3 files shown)


wanderingstan:
This looks great! Didn't realize we could query all past events on a contract like that. Very useful.

wanderingstan:
@tyleryasaka @joshfraser  

This commit is to origin-js _and_ has changes to the contracts, so it would require a full deploy and re-migration of old data. 

Just wondering if we want to do that. If I have time later tonight I'll start the process off of develop and see if I hit any snags. But I think we should be prepared to not include this branch if it gets hairy.

PR #216, opened 2018-05-29
Filter out image URLs that are not `data:` or `dweb:` protocol

wanderingstan:
### Checklist:

### Description of the problem being solved

Filter out image URLs that are not `data:` or `dweb:` protocol. This is first step to linking to image IPFS hashes in our Listing json. 

Currently we filter out non-`data:` URLs in the Dapp. 

I think origin-js is the right place to filter, though an argument could be made that DApp developers should be free to take on risks of accepting URLs with unsafe protocols (eg `http`)

Next steps would be:
- Remove URL checks in DApp
- Change listing submission code to upload images first, then link to them in listing json with links like  `dweb:/ipfs/QmWP28bNAJbkiKrXHAHzotKCvLyNragErycSYQQR9KiFby`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17200,8 +17200,7 @@
     "requires-port": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/requires-port/-/requires-port-1.0.0.tgz",
-      "integrity": "sha1-kl0mAdOaxIXgkc8NpcbmlNw9yv8=",
-      "dev": true
+      "integrity": "sha1-kl0mAdOaxIXgkc8NpcbmlNw9yv8="
     },
     "resolve": {
       "version": "1.5.0",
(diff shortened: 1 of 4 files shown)


wanderingstan:
Actually...hold off on merging this. I'm re-thinking the right way to do this: perhaps should just be origin-js helper function that the dapp uses, which re-writes dweb uris into URLs using our gateway, _and_ filters out unsafe URLs.

DanielVF:
I like having any checking/rewriting being called by origin.js, rather than optionally by the dapp because:

- It removes a possible class of security errors that dapp developers could easily make.
- By enforcing the defaults to IPFS and data, we ensure more interoperability among contracts/DAPP/aggerigators.

If we/others need other need other protocols for images, we could potentially add that as a config option to origin.js.

micahalcorn:
Closing in favor of #320 🤔

PR #222, opened 2018-06-01
Client-side notifications

### LINKED ISSUE #211: Client-Side Notifications
Tyler and I have been discussing how to handle notifications in the DApp.

**Assumptions**
- All notifications will be derived from existing blockchain transactions, [such as purchase state changes or other timestamps](https://github.com/OriginProtocol/bridge-server/issues/6#issuecomment-384096004).
- No identifying information should be stored on a bridge server (such as an auth token).
- DApp[s] UX will require read/unread status to be tracked.
- Users and/or DApp developers may change their bridge server URL from time to time.
- Users may interact with Origin.js using the same wallet/identity but different devices.

**Recommendation**
- We will not store anything related to notifications in the bridge server, including sent/read timestamps.
- Origin.js will manage the state of client-side notifications. It will provide two methods for DApp use: "mark as read/unread" and "get all notifications".
- We will use [Store.js](https://github.com/marcuswestin/store.js/) or a similar library for cross-browser support of local storage.
- When origin-js is first accessed from a device, a timestamp will be recorded in local storage. Only notifications that "occur" after this timestamp will be shown as "unread" on this device. All prior notifications will be treated as if they have been read.
- We will need some way to predictably identify the notification within Origin.js, perhaps using something like the triggering block number plus some other identifying scheme.

cc @crazybuster

tyleryasaka:
### Checklist:

### Description:

This creates an API for client-side notifications with no dependencies other than local storage. Notifications are built directly from blockchain data, and read/unread state is stored in local storage.

I implemented the following notification types:
- For seller:
	- `seller_listing_purchased`
	- `seller_review_received`
- For buyer:
	- `buyer_listing_shipped`
	- `buyer_review_received`

This can be adjusted/expanded in the future as needed.

Closes #211


micahalcorn:
Thanks, @tyleryasaka 👏 Really great work.

I am getting two notifications with the same `id` when I call `notifications.all` from account `0x627306090abaB3A6e1400e9345bC60c78a8BEf57`. The duplicate notification is as follows:

```
{
  id: "seller_listing_purchased_0x23172349165f59c95567e1e993e473ba914d0dd4fe59ee398b051cfda4cea622",
  type: "seller_listing_purchased",
  status: "read"
}
```

Also, I'm going to need to map the role of the user to this object in addition to obviously getting the relevant data (listing details, purchase details, etc) to produce a friendly message. Should any of this be done on the origin-js side since you already seem to have strung the necessary async calls together?

package.json
```diff
@@ -38,6 +38,7 @@
     "form-data": "^2.3.2",
     "map-cache": "^0.2.2",
     "rlp": "^2.0.0",
+    "store": "^2.0.12",
```
micahalcorn: 👍

micahalcorn:
@tyleryasaka here is my WIP branch: https://github.com/OriginProtocol/demo-dapp/tree/notifications

Since we aren't going to be "indexing notifications" on the bridge server, then I'm wondering what is the best way to limit the number of requests for this information. Perhaps it makes sense to put all of a user's listings and purchases in the DApp redux store as soon as they load the app. Then we can probably map all of my content described above using the stored collections. That would effectively cache all of the other routes too. ❓

PR #223, opened 2018-06-01
Disallow purchase of listing by its seller + test helper

### LINKED ISSUE #213: Prevent seller from purchasing own listing
In the past, we have experienced some unexpected behavior with reviews when a seller and a buyer are the same account. [The decision was recently made to prevent DApp users from purchasing their own listings](https://github.com/OriginProtocol/demo-dapp/pull/232/commits/b495d9c477cefa344c45892c0f1f757461b0eca5). There may be some instances where a user wants to cause a listing to no longer be available, but this should be accomplished by "closing" the listing. A user with a large inventory of listing might prefer to show some as sold, perhaps as a result of selling them off-chain, but I'm not convinced that allowing a seller to be a buyer of the same listing is the appropriate way to handle this case. Unless there is an argument for allowing it, let's prevent this behavior at the contract level.

tyleryasaka:
### Checklist:

### Description:

I branched off of https://github.com/OriginProtocol/origin-js/pull/221 and implemented an alternate solution for executing test calls in the context of a specific active account.

This approach is inspired by @cuongdo 's approach and is quite similar, but I prefer this approach because it uses a built-in web3 property `eth.defaultAccount` (that we should probably be using anyways), rather than adding our own property for monkey-patching the current account during tests. It also places all the responsibility of reseting the active account in the test helper, keeping the test code clean and easier to maintain.

Closes #213

### DIFF
--- a/src/services/contract-service.js
+++ b/src/services/contract-service.js
@@ -75,7 +75,7 @@ class ContractService {
   }
 
   // Returns the first account listed, unless a default account has been set
-  // expcicitly
+  // explicitly
   async currentAccount() {
     const accounts = await this.web3.eth.getAccounts()
     const defaultAccount = this.web3.eth.defaultAccount
--- a/src/services/contract-service.js
+++ b/src/services/contract-service.js
@@ -74,10 +74,12 @@ class ContractService {
     return hashStr
   }
 
-  // Returns the first account listed
+  // Returns the first account listed, unless a default account has been set
+  // expcicitly
   async currentAccount() {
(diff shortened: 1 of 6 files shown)


src/services/contract-service.js
```diff
@@ -74,10 +74,12 @@ class ContractService {
     return hashStr
   }
 
-  // Returns the first account listed
+  // Returns the first account listed, unless a default account has been set
+  // expcicitly
```
cuongdo: nit: typo

tyleryasaka: fixed

PR #225, opened 2018-06-05
Use eslint for finer control of code format

tyleryasaka:
### Checklist:

### Description:

Switches us from using just [prettier](https://github.com/prettier/prettier) to using [eslint + prettier](https://github.com/prettier/prettier-eslint-cli)

I switched to eslint because it is much more configurable, and I have had a great experience it with it on another project.

We now have:
- Enforced consistency over var/const/let. Followed style suggested in [this article](https://medium.com/javascript-scene/javascript-es6-var-let-or-const-ba58b8dcde75). Also see https://github.com/OriginProtocol/origin-js/issues/192.
- Consistent use of quotes (~double~ single or backticks)
- Lots of other nice restrictions that are enabled by eslint out of the box, like preventing unused vars


.eslintrc.js
```diff
@@ -18,8 +18,16 @@ module.exports = {
...
   rules: {
-    'no-console': ['off'],
-    'prefer-const': ['error'],
-    'semi': ['error', 'never']
+    "no-console": ["off"],
+    "prefer-const": ["error"],
+    semi: ["error", "never"],
+    quotes: [
+      "error",
+      "double",
```
micahalcorn: double quotes 😫

contracts/test/TestListingsRegistry.js
```diff
@@ -38,14 +31,14 @@ contract("ListingsRegistry", accounts => {
...
+    const listingCount = await instance.listingsLength()
     assert.equal(
       listingCount,
       initialListingsLength + 1,
       "listings count has incremented"
     )
-    let [
-      listingAddress,
+    const [
+      ,
```
micahalcorn: Are these now rogue commas 😕

contracts/test/TestPurchase.js
```diff
@@ -32,8 +32,8 @@ const AWAITING_PAYMENT = 0 // Buyer hasn't paid full amount yet
 const SHIPPING_PENDING = 1 // Buyer hasn't paid full amount yet
 const BUYER_PENDING = 2 // Waiting for buyer to confirm receipt
 const SELLER_PENDING = 3 // Waiting for seller to confirm all is good
-const IN_DISPUTE = 4 // We are in a dispute
-const REVIEW_PERIOD = 5 // Time for reviews (only when transaction did not go through)
+// const IN_DISPUTE = 4 // We are in a dispute
+// const REVIEW_PERIOD = 5 // Time for reviews (only when transaction did not go through)
```
micahalcorn: Reviews ended being parts of 2️⃣and 3️⃣. Should this just be removed completely, or are we planning to bring back a post-dispute review period @DanielVF?

micahalcorn:
I'll defer to @DanielVF on some of this. I certainly favor `const` > `let` > `var` and generally eslint. 👍

jordajm:
I really like eslint too @tyleryasaka - thanks for working on this. 

One suggestion would be to consider using the `jsx-quotes` rule to enforce use of double quotes in JSX attributes (which is similar of HTML, where double quoting is more conventional) while staying with single quotes in the JavaScript (which seems to be the standard throughout the app so far, and is definitely my personal preference)
https://eslint.org/docs/rules/jsx-quotes
Thoughts on that?

contracts/test/TestListingsRegistry.js
tyleryasaka: @micahalcorn Nah, it's just destructuring syntax for ignoring values we don't care about. See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment (section "Ignoring some returned values")

.eslintrc.js
tyleryasaka: @micahalcorn Do you prefer single quotes? I'm not opinionated on it, I just chose double because more of our code was using double. 😃

contracts/test/TestPurchase.js
tyleryasaka: Imho we don't have to make that decision now - that's why I just commented them out, so we don't have to deal with it in this PR. (I think it's better to just keep this focused on the linter change.)

tyleryasaka:
@jordajm Once this gets merged I can do a similar PR in demo dapp repo!

.eslintrc.js
```diff
@@ -0,0 +1,35 @@
+module.exports = {
```
nick: Can we put this in package.json instead of another dotfile?

package.json
```diff
@@ -62,7 +60,6 @@
     "mocha": "^5.1.1",
     "mocha-loader": "^1.1.3",
     "node-watch": "^0.5.8",
-    "prettier": "1.12.1",
```
nick: Does eslint do automatic file formatting? In other projects I've used both prettier and eslint together without issues

.eslintrc.js
nick: I vote for single quotes too :-)

.eslintrc.js
nick: eg https://github.com/OriginProtocol/identity-playground/blob/master/package.json#L105

package.json
tyleryasaka: @nick Yeah eslint does file formatting. Eslint does everything we need it to do - I don't see any use for prettier.

nick: Apparently prettier is [more powerful](https://github.com/prettier/prettier-eslint/issues/101) than eslint at formatting files. Perhaps we can run both eslint and prettier separately on demo-dapp, see which does a better job and decide to dump prettier based on that?

DanielVF:
I prefer const > let > var as well.

`const` does catch errors from time to time. I don't like it, since it makes half my lines longer than the should be and harder to read, but the utility is undoubtable, and so I use it anyway.

Prettier as done a great job of just automatically working and bringing the code into formatting spec. There a big difference between a tool you just run and it works, vs something that spits out lists of spacing errors that you have to go fix yourself - which has always been my past experience with eslint. (Has it gotten better?)

ambertch:
RE: single / double quotes, a convention I have heard of is to use double
quotes to specifically denote strings, since a lot of times in code there
will be string types that are not necessarily strings.

Sort of like the string and symbol distinction in ruby. They can be
substituted for one another, though a symbol usually denotes a string value
as opposed to a string.

.eslintrc.js
tyleryasaka: done

tyleryasaka:
Updated to use [prettier-eslint-cli](https://github.com/prettier/prettier-eslint-cli)

tyleryasaka:
Single quotes it is! The vote is 2:1 in favor of single quotes. (I abstained but if I had to pick I also like single quotes - easier for me to read and type.)

tyleryasaka:
Next up: spaces vs tabs... jk jk

tyleryasaka:
@micahalcorn I believe I have addressed all the concerns raised here. How do you feel about merging this? (Good to merge as quickly as possible to reduce conflicts.)

PR #227, opened 2018-06-06
Dev docker container 

ambertch:
### Checklist:

### Description:

Adds docker container under /devcontainer directory, running:
  * ubuntu 16.04
  * python 3.6
  * bridge server (develop)
  * postgres 9.3
  * redis 4.0
  * origin-js (develop) for ganache, ipfs
  * elasticsearch 5.4.3


devcontainer/files/scripts/start-celery.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+source $BRIDGE_SERVER_PATH/bridge-server-venv/bin/activate
+celery -A util.tasks worker -c=1 -E -B -l debug
```
franckc: Should the celery beat process (i.e. the scheduler) also get started here ? Something along the line of:
`celery -A util.tasks beat`

micahalcorn: Yes, please include all things celery 🌵

devcontainer/Dockerfile
```diff
@@ -0,0 +1,117 @@
...
+RUN curl -sL https://deb.nodesource.com/setup_9.x | bash -
+RUN apt-get install -y nodejs
+
+# IPFS, Ganache (through origin-js)
+WORKDIR /opt
+RUN git clone https://github.com/OriginProtocol/origin-js.git
+WORKDIR /opt/origin-js
+RUN npm run install:dev
+
+# Postgresql 9.3
```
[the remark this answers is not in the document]
ambertch: Thanks for the observation. 9.3 is the version referenced in the bridge server README (https://github.com/OriginProtocol/bridge-server/blob/master/README.md)

Going off what was done for Elasticsearch (running the version closest to production), I see that Heroku Postgres defaults to 10 https://devcenter.heroku.com/articles/heroku-postgresql#version-support-and-legacy-infrastructure

I'll update to that version.

devcontainer/files/scripts/start-celery.sh
ambertch: Gotcha. This change was made when I was setting the logging level and saw that the -B argument runs celery beat with the worker process. It's meant for dev environments, and allows one tab or process to be run instead of two.

Along the lines of mirroring production, I see that Celery can be run this way in Heroku: https://drdaeman.github.io/heroku-djcelery-example/

@Gzing - in our production bridge server, are we running a two dyno, three dyno, or add-on setup for celery? Thanks!

tyleryasaka:
So @ambertch and I were confused by some interesting behavior here. We realized that origin-js is being built in the docker container, but the demo dapp (when linked to the local origin-js) is importing build files that are out of date (the up-do-date ones are isolated inside the container). We were both very confused how the demo dapp was determining the deployed contract addresses (which it was doing with no issues).

I finally figured it out. I didn't realize this, but the addresses that get deployed during origin-js local build are the same every time, as long as you're deploying the same contracts in the same order. (Now that I think about it it makes sense. The address where a contract will be deployed is deterministic, based on the sender's address and the "nonce", which is just the number of transactions from that address. This is how we're pre-signing attestations for ERC725/735.)

In other words, the docker container running origin-js being out of sync with the ABI that your local demo dapp is using *happens* to work just fine right  now - but if you are doing local development and the contracts deployed during migrations change at all, this may no longer work. So it may work a lot of the time, but it's not reliable. We really need to be mounting the build directory into the docker container so that the demo dapp is using most up-to-date ABI. If we want to be able to use this for local development.

That said, this code is isolated from the rest of our codebase, so I will go ahead and merge this. We can create further issues and improve on what we have from there. Thanks @ambertch for getting this kicked off!

wanderingstan:
Maybe dumb question on the docker setup: When we say the "docker container is running origin-js", what exactly does that mean? Origin-js is an npm package that just gets included in other projects, not run directly.

My first thought was that it means that the dapp (running on laptop) is `npm link`-ed to the version of origin-js in the docker container. But that seems impossible, as one can't make a symlink to another machine. What am I missing?

tyleryasaka:
@wanderingstan origin-js does get run directly for local development; it runs local IPFS/blockchain and runs the initial migrations to deploy the contracts. That's what we mean when we say the docker container is "running" it.

@wanderingstan Yeah the demo dapp will work if symlinked to your local origin-js *source code* - and then our docker command maps the ports correctly, so that the demo dapp connects to the running blockchain/IPFS in docker. As far as getting the correct contract addresses in the ABIs; that is what my previous comment was discussing. It happens to work right now, until we make changes to our migrations (because the contracts get deployed in the same order and thus have the same address every time).

It's a bit of an awkward setup and can be improved. I'm creating issues in github right so that we can make this better.

Does that answer your question?

wanderingstan:
@tyleryasaka All clear now. For clarity in the future I would refer to this as "The docker container is running the origin-js dev environment."  In my thinking, origin-js is just the npm package, and all the rest (IFPS, local blockchain, contracts) are things that it interacts with. 

Another maybe dumb question: Why do we need to run the origin-js dev environment in the docker container? IIRC, the initial impetus for the docker container was just to run the bridge server (with its many dependencies). As @ambertch pointed out, docker containers aren't meant to run a lot of programs. Is there some reason we can't have IPFS/blockchain running on the local machine and just bridge server in the container? (Again apologies, I know I've been out of the loop on this--just understanding our architecture.)

tyleryasaka:
@wanderingstan Got it 👍. Yeah it's a bit confusing because our origin-js repo is more than just JS; it also contains our contracts, and has become our natural "dev environment" for running the blockchain and IPFS. I have to stop and think about it to keep it all straight in my head. 😄 

@wanderingstan Maybe we don't need to run origin-js in docker. I think we still have a lot of clarification to do around the purpose of the container.

Although if we do want to run lots of processes, there is [docker compose](https://github.com/OriginProtocol/origin-js/issues/229) - might be worth considering.

Created an issue so we can continue discussing: https://github.com/OriginProtocol/origin-js/issues/230

PR #232, opened 2018-06-08
Tyleryasaka/dev container tweaks

tyleryasaka:
### Checklist:

### Description:

Makes some modifications to our docker container setup. Thanks again @ambertch for all the work getting it to this point. Made this easy for me.

- Moves management of ganache/ipfs from container. This can already be done with a single command outside the container (`npm run start`), so no need to complicate things by putting it in the container. (My bad - I think it was originally my suggestion to put it in there.) See https://github.com/OriginProtocol/origin-js/issues/230. Also, if this is merged, https://github.com/OriginProtocol/origin-js/issues/228 will be unnecessary.
- Creates npm commands to easily manage the docker container. See https://github.com/OriginProtocol/origin-js/issues/231
- Renames the directory to `bridge-container`, as the container is now just managing the bridge server.


cuongdo:
This escaped me when I reviewed it, but would this more logically belong in `bridge-server`?

ambertch:
A side effect of running the blockchain outside of the container is that the bridge server in the container can't reach http://localhost::8545 (in the container run: pm2 logs celery)...

What do you think about mounting origin-js as a volume? That way, the addresses of any deployed contracts will still be kept in sync between origin-js, the dapp linking it, and the container. There might also be docker networking configuration that does this.

tyleryasaka:
@ambertch Oh right. Hmm, yeah mounting as a volume seems like the best option then.

The other advantage of mounting as a volume is that we could add a command to symlink the [contract ABIs on the bridge server](https://github.com/OriginProtocol/bridge-server/tree/master/contracts) with the origin-js contract ABIs (in the `contracts/build` directory). Then if you were doing solidity development the bridge server would automatically get any ABI changes.

tyleryasaka:
@cuongdo There's a few different ways we could set this up. I think it all comes down to the use cases. In this case, I think this container exists for people that are working on origin-js (not bridge-server), so it's a bit more convenient for it to exist in this repo. People can just start hacking away at origin-js without having to clone yet another repo. That said, it might be a bit cleaner to keep the bridge-server container in, well, the bridge-server repo.

The other use case that comes to mind is for people doing *bridge-server* development. I'm still thinking about how to best support that.

Some options in my head are:
1. Keep this container in origin-js, for the purpose of making origin-js development as easy as possible
1. Place this container in bridge-server, which is cleaner and probably feels like the more "correct" place to put it.
1. Try implementing this "Origin-in-a-box" idea that has been floated around; I'm visualizing this as its own repository that uses docker-compose to clone all of our different repos (demo-dapp, origin-js, bridge-server) and spin them up into separate containers that interact with each other, and mount them as volumes so that you can do development on any or all of the repos.

The first 2 options don't address the issue of bridge-server development that needs to have origin-js running to support it. For that use case, we have a couple options:
1. Just ask the user to symlink the origin-js and bridge-server contract ABI directories
1. Use pip to more elegantly manage this symlinking (npm has this capability; I'm not familiar enough with pip to know if it supports this)
1. Place a docker container inside the bridge-server repo that spins up origin-js

I'm sort of leaning towards the origin-in-a-box idea. If we can get it to work, it could be really awesome. It would be a single place to easily spin up all of our different pieces and see them interact with each other. And it would take care of all the development use cases at once, with no redundancies and weird container placement schemes, or asking the user to have 3+ terminal tabs open and running several commands in each.

cuongdo:
I also lean toward the Origin-in-a-box idea with, if possible, the ability to swap out the in-container Git repos with Docker volumes that mount the repos from the host. This would allow editing of source using a GUI editor on the host but testing in an easy to setup Docker environment. I'm no expert on Docker, so I'm not sure what challenges exist here.

tyleryasaka:
@cuongdo I'm also not a docker expert but yes I'm thinking along the same lines; I think mounted volumes would be necessary for this to be useful in development, for using your own GUI code editor like you say.

My goal is to do some exploration on this idea today.

Issue #239, opened 2018-06-12
Exception being swallowed in contract service


micahalcorn:
[This error](https://github.com/OriginProtocol/origin-js/blob/master/src/services/contract-service.js#L197) is not being surfaced to consumers of the `getListing` method. It should probably be thrown in addition to a friendly `console.log` [like this](https://github.com/OriginProtocol/origin-js/blob/master/src/services/contract-service.js#L123) or we should develop a more consistent strategy for providing details about the error.

marcogbarcellos:
Hi @micahalcorn I can get this one!

micahalcorn:
Great, thx @marcogbarcellos 👍

marcogbarcellos:
#242

PR #241, opened 2018-06-13
Get listing error

marcogbarcellos:
Small PR regarding the issue #239 .

- Using `console.log(STRING)` and `throw error` instead of `throw new ERROR(String)`
- updating unit test
- Minor tweak: removing unnecessary template strings to use single quotes instead.


micahalcorn:
@marcogbarcellos since you are now a contributor in this repo, I recommend branching from [develop](https://github.com/OriginProtocol/origin-js/tree/develop), adding your fix and then pushing your branch here to the official repo. You won't need to use a fork, just point your PR to develop instead of master. 👍

marcogbarcellos:
Sounds good @micahalcorn , will do it. thanks.

Issue #250, opened 2018-06-21
Listing IPFS hash versioning


tyleryasaka:
In our [fractional usage discussion](https://github.com/OriginProtocol/origin-js/issues/146), we've decided that the IPFS hashes should be updatable and versioned. Since we're going to have a base listing contract, I think the versioning should happen there (whether our dapp supports IPFS data updates for other listing types or not).

I think the simplest way to implement this right now is to have an array of IPFS hashes on the base listing contract. The last item of the array is considered the most recent version. ~We could eventually get fancier and put timestamps on the versions, but I don't think that's necessary yet.~ EDIT: I think we should go ahead and be fancy and add timestamps. The reason we're adding versioning in the first place is primarily arbitration. Timestamps seem pretty essential to arbitration and also go naturally with versions, so we should probably go ahead and include them.

nick:
I think we should try to avoid storing anything that can be determined by other methods so we're not using contract storage unless absolutely necessary. Using a single value instead of an array for ipfsHash will also reduce contract complexity. 

Instead, we could emit a ListingUpdated(bytes32 ipfsHash) event whenever a listing is updated. We could then use the [getPastEvents](http://web3js.readthedocs.io/en/1.0/web3-eth-contract.html#getpastevents) method from web3 filtering on ListingUpdated to give us the ipfsHash history. The returned events also have a blocknumber property which we can use with [getBlock](http://web3js.readthedocs.io/en/1.0/web3-eth.html#getblock) to determine the timestamp.

tyleryasaka:
@nick Fair point.

That said, I realized that if we make listings editable, we'll need to deal with this issue: https://github.com/OriginProtocol/origin-js/issues/19

TLDR: purchase contracts should reference specific versions of listing contracts, if listing contracts are editable.

We can still rely on events to get previous IPFS hashes. However, a couple of concerns:
1. If at any point in the future we want anything editable to be stored directly on-chain (as opposed to in IPFS), then we'll need to store versions explicitly on-chain.
1. I wonder if it could be ambiguous or confusing which IPFS hash a purchase contract points to, if we're just looking at the block number of the event? If we stored versions on-chain, we could have the purchase contract programmatically save the version number of the listing that it points to. If we're just looking at events, we'll have to infer what the active version of the listing was at the time of purchase contract creation.

With those thoughts in mind, I'm not sure it's worth trying to optimize this part right now. I would think the first step towards optimization would be moving purchase contract code to a library.

[...]

Issue #249, opened 2018-06-20
Create FractionalListing contract


tyleryasaka:
We should create a first iteration of our fractional usage listing contract. The contract should inherit from the base `Listing` contract described in https://github.com/OriginProtocol/origin-js/issues/248.

tyleryasaka:
Done via #251

Issue #248, opened 2018-06-20
Refactor listing contracts


tyleryasaka:
Our listing contract should be refactored so that we have a base `Listing` contract, and an inherited `UnitListing` (which does what our current `Listing` contract does).

tyleryasaka:
Yes, this is done 👍

Issue #146, opened 2018-05-02
🎓Fractional usage and calendaring


micahalcorn:
The eventual goal is to support fractional-use DApps that will involved a calendaring function. Some ideas around this are outlined in the [white paper](https://www.originprotocol.com/en/whitepaper).

**Things to Consider**
- mapping user-friendly times and smart contracts
- challenges with daylight savings, time zones, etc
- business process requirements
- [iCalendar standard](https://tools.ietf.org/html/rfc5545)
- variable pricing on-chain in an efficient format

**To Do**
- [x] UI changes supporting time intervals ranging from minutes to year(s)

DanielVF:
One way that is flexible, simple, and gas cheap:

We could have N times slots. Each timeslot has a start and and time in UTC seconds (unix timestamp). The listing IPFS json has an array of these slots. The listing contract has an availability (bytes or array of numbers) as a bit string. We can do cheap bitwise operations on these to check availability or mark as used.

joshfraser:
I love the idea of utilizing bit masks.

A couple things to remember, we'll eventually want the ability to set variable pricing for different time slots (we've been calling them "intervals") and we'll need some way to track who the purchaser of each Interval was, not just availability (although maybe we get that already from the purchase contract?).

It may also simplify things if each Interval is a fixed length of time. That allows you to have simple rules like every 6th and 7th interval should have a higher price (say for weekend pricing).

DanielVF:
Are there any changes to the purchase flow that we need for this fractional usage sprint?

nick:
I found some resources that may or may not be useful for this discussion:

- [Interval Tree implementation in Solidity](https://github.com/gnidan/interval-trees-solidity)
- [Timely Resource smart contract design](https://docs.google.com/document/d/11WCubnkejEs-1K_I17DgQSlNtGJO8TS2vXkTCijytGg/edit)

joshfraser:
Thanks for kicking this off Daniel and enumerating the various decisions we need to make.

While different types of listings should be able to have variable interval sizes (or slots as you call them), I think we should have a fixed interval size for a specific listing. Things like being able to set a minimum number of intervals (ie. min 2 nights booking) is pretty important and easier to reason about when the intervals are a consistent size.

wanderingstan:
[...]

PR #251, opened 2018-06-21
Fractional usage contracts v1

tyleryasaka:
### Checklist:

### Description:

This is a first stab at supporting fractional usage listings in our smart contracts.

What's going on here:
- Refactors listing contract structure so that we now have a base `Listing` contract and inherited `UnitListing` and `FractionalListing` contracts. `UnitListing` does what our current `Listing` is doing.
- Refactors `Purchase` contract to work with both `UnitListing` and `FractionalListing` contracts. We may eventually want to do something similar with the `Purchase` contract, where we have a base and inherited contracts for variations between listing types. However, `UnitListing` and `FractionalListing` purchases are similar enough right now that I decided to just have one `Purchase` contract for simplicity.
- Fleshes out basic functionality to support fractional usage listings. Right now we're assuming all price and availability data will be stored in IPFS, and confirmed manually by the seller (rather than being enforced on-chain). 
- Implements a versioning system for fractional usage listings. We're allowing the ipfs hash to be updated, but only in the sense that the seller can release new "versions" of the ipfs hash. This allows the listing to be edited without affecting existing purchases. (All purchases now store a version number in addition to a listing contract address.)

Relevant issues:
- 🎓Fractional usage and calendaring #146
- Refactor listing contracts #248
- Create FractionalListing contract #249
- Listing IPFS hash versioning #250
- Think through listing contract mutability vs. immutability #19


cuongdo:
Good start, @tyleryasaka.

PR #272, opened 2018-07-11
Jordajm/fractional schema tweaks

tyleryasaka:
### Checklist:

### Description:

Bug fixes and schema tweaks

### DIFF
--- a/contracts/contracts/Purchase.sol
+++ b/contracts/contracts/Purchase.sol
@@ -90,8 +90,8 @@ contract Purchase {
   function data()
   public
   view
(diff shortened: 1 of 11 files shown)


tyleryasaka:
Haven't confirmed this theory but I think what might be happening here is that we're creating too many listings/purchases during are tests, and that is causing the notifications tests to time out because there are too many things to load. (Issue is described here: https://github.com/OriginProtocol/origin-js/issues/245.)

Will investigate

tyleryasaka:
Old theory appears to be incorrect. Fixed the test issue in https://github.com/OriginProtocol/origin-js/pull/273 but that doesn't fix these failures.

PR #276, opened 2018-07-19
Add parameters required for Twilio Verify API

tomlinton:
### Checklist:

### Description:

- Add `country_calling_code` parameter. Previously `phone` was a concatenation of this and the local phone number, but the Twilio Verify API requires them separate.
- Add `method` because we can support 'call' or 'sms'. This can be used to give people a choice in origin-dapp.
- Add `locale`. The Twilio Verify API uses a sensible default for the verification language based off the country of the phone number. We can override it though, so in origin-dapp we could use the language the user has selected.

Docs are sitting in origin-docs on the branch with the same name.

### DIFF
--- a/src/resources/attestations.js
+++ b/src/resources/attestations.js
@@ -58,16 +58,22 @@ class Attestations {
     }
   }
 
-  async phoneGenerateCode({ phone }) {
-    return await this.post('phone/generate-code', { phone })
+  async phoneGenerateCode({ country_calling_code, phone, method, locale }) {
+    return await this.post('phone/generate-code', {
+      country_calling_code,
+      phone,
+      method,
+      locale
+    })
   }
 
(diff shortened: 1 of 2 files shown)


src/resources/attestations.js
```diff
@@ -58,16 +58,22 @@ class Attestations {
     }
   }
 
-  async phoneGenerateCode({ phone }) {
-    return await this.post('phone/generate-code', { phone })
+  async phoneGenerateCode({ country_calling_code, phone, method, locale }) {
```
franckc: Could you document this interface change by submitting a PR in our [origin-docs repo](https://github.com/OriginProtocol/origin-docs) to update the [attestation APIs](https://github.com/OriginProtocol/origin-docs) ?

Same comment for the method phoneVerify below.

Thanks in advance !

PR #285, opened 2018-07-24
Refactor blockchain transactions and support callbacks

micahalcorn:

### Checklist:

- [ ] Update any relevant READMEs and [docs](https://github.com/OriginProtocol/docs) 🤙 **WILL DO**

### Description:

In an effort to enable https://github.com/OriginProtocol/origin-dapp/issues/251...

- Removes the extra `contractFn` on the `ResourceBase` and updates `Purchases` to use the one in `contractService`
- Updates the comment describing `contractFn`
- Removes support for `waitTransactionFinished`, which arbitrarily considered a first block confirmation to equate to "finished" using an interval to poll for a block number increment
- Removes support for the `whenFinished` promise
- Adds support for an optional callback function that is called on each [block confirmation event](https://web3js.readthedocs.io/en/1.0/callbacks-promises-events.html#promievent)
- Adds a `gas` argument of `100000` for the `buyerConfirmReceipt` method
- Changes the return value of `contractFn` from a raw transaction receipt to a custom object containing two properties: the transaction receipt and the timestamp for the relevant block


micahalcorn:
Thx @tyleryasaka, good to know. Is that because the value may end up not being an object, or why?

tyleryasaka:
@micahalcorn Nah, it's just that the word "object" tells you what format the data is in, not what the data represents. I just prefer to use variable names that describe what the data is actually doing. If for example you find yourself having 2 variables called `fooBar` and `fooBarObject`, that's a sign that you might be representing data in a weird way, or not describing it well.

In this case I would try to find a more meaningful distinction between the returned object from origin-js method calls and the actual web3 transaction receipt that is returned as a property. I might call the returned object `result` and the transaction receipt `web3TransactionReceipt`, but that's just one idea.

PR #291, opened 2018-07-30
Upgradability

tyleryasaka:
### Checklist:

### Description:

Not ready to merge yet. Waiting on dapp updates. Rewrites a lot of origin-js to include an upgradability strategy as well as other architectural changes.


nick:
This is a fair amount to review :) - perhaps we could get on a call sometime to walk through the code?

contracts/contracts/EvolvingRegistry.sol
```diff
@@ -0,0 +1,85 @@
...
+
+  function size() public constant returns (uint256) {
+    return entries.length;
+  }
+
+  function addEntry() public returns (uint256) {
+    // This lookup will succeed and return 0 for a contract address that we don't have an entry for...
+    uint16 _entryTypeIndex = entryTypesByAddress[msg.sender];
+    // ...which will give us a vaild, but wrong entry type.
+    // Therefore we check that we have actualy looked up the matching entry type
```
cuongdo: nit: typo

contracts/contracts/EvolvingRegistry.sol
```diff
@@ -0,0 +1,85 @@
...
+    The value of each entry is the index of the entry type.
+    The index can be used to get the entry type contract.
+    The entry type contract handles all logic for the entry.
+  */
+  uint16[] public entries;
+
+  function addEntryType(address _contractAddress, string _name) public onlyOwner() {
+    // Don't allow the same contract address to be added twice
+    uint16 trialIndex = entryTypesByAddress[_contractAddress];
+    require(entryTypes[trialIndex].contractAddress != _contractAddress);
```
cuongdo: Maybe add `require(_contractAddress != 0)`? `addEntry()` seems to depend on `entryTypesByAddress[0]` being 0.

contracts/contracts/EvolvingRegistry.sol
```diff
@@ -0,0 +1,85 @@
...
+    uint16 entryTypeIndex = entryTypeLength;
+    entryTypes[entryTypeIndex] = EntryType(_contractAddress, _name, true);
+    entryTypesByAddress[_contractAddress] = entryTypeIndex;
+    entryTypeLength++;
```
cuongdo: There's no check for overflow here. Maybe use `SafeMath.add()` instead?

src/adapters/v01/listings_adapter.js
```diff
@@ -0,0 +1,372 @@
...
+  async getPurchase(listingIndex, purchaseIndex) {
+    return await getPurchaseHelper(
+      this.contractService,
+      this.ipfsService,
+      listingIndex,
+      purchaseIndex
+    )
+  }
+
+  async getPurchases(listingIndex) {
```
cuongdo: Maybe add a TODO about this not scaling once there are many purchases because of memory usage? I'm not sure about the memory limits of in-browser JS, but there's probably some practical limit.

test/resource_reviews.test.js
```diff
@@ -0,0 +1,99 @@
...
+    purchaseIndex
+
+  beforeEach(async () => {
+    const ipfsService = new IpfsService({
+      ipfsDomain: '127.0.0.1',
+      ipfsApiPort: '5002',
+      ipfsGatewayPort: '8080',
+      ipfsGatewayProtocol: 'http'
+    })
+    const provider = new Web3.providers.HttpProvider('http://localhost:8545')
```
cuongdo: not specific to this PR, but we should consider using ephemeral ports when binding services during tests. this helps avoid port conflicts between a local development blockchain and test blockchains

cuongdo: note that this is not for this PR... just noting

nick:
I'm not sure what the benefit of having the Evolving Registry concept on-chain is. From what I can see, the idea is to have a unique listing ID that works across different contracts. I see how that would be useful from an OriginJS or dapp perspective, but not sure what the benefit is at the contract level.

Instead, for the purposes of OriginJS and the Dapp, couldn't we just split the listing ID into two pieces, where the first is the contract version and the second is the listing ID on that contract? For example, listing 001-289 would be listing 289 on contract version 1. 003-97 would be listing 97 on contract version 3.

That way we can remove the extra gas, complexity and security implications of an Evolving Registry contract while keeping the ability to address listings on different marketplace contracts.

contracts/contracts/listings/v01/Listings.sol
```diff
@@ -0,0 +1,206 @@
+pragma solidity 0.4.23;
+
+import "../../EvolvingRegistry.sol";
+import "./Escrow.sol";
+
+contract V01_Listings {
+  event ListingChange(uint256 indexed _listingIndex, bytes32 _ipfsHash);
```
DanielVF: The `_listingIndex` in this event is the same as the `_entryIndex` in the registry events. Perhaps since are the same number, they should be called by the same name.

tyleryasaka:
Appreciate all the feedback - I'll be looking over it. Just a heads up, this is a work in progress so I will be pushing commits to this branch. Created the PR for better visibility on what I'm doing here.

tyleryasaka:
@nick Wish we'd looped you into this discussion earlier!

Yeah I like the idea you mentioned. I had also considered keeping track of various listing contracts entirely off-chain.

I went with this approach to be on the safe side of having a master registry on-chain. The main advantage I could see is being able to implement on-chain logic that needs to know whether a listing is an Origin listing or not.

For example, token incentives related to listings.

But yeah this may be over-engineered and overkill - felt like the safer option to me. We can always move from having a master on-chain registry to not having one, but it's harder to move in the other direction.

I could be persuaded either way. Any other thoughts? @DanielVF @cuongdo ?

cuongdo:
I did a cursory review, because a lot of stuff is still in flux. Here are my high-level comments:

1. I'm curious what the overhead of scanning a bunch of logs is. For example, in `getPurchaseLogs`, you can from block 0. That might take a lot of time on mainnet. We could bound the search from the block number at which the the purchase started. This could be stored in IPFS. (This can be done later.)
2. The structure of the adapter looks generally good. I like the encapsulation of interacting with the contracts there. It doesn't seem particularly challenging to have either @tyleryasaka's contracts or @nick's contracts with this in place.
3. I think only allowing 2 digits for `V01ListingAdapter`, etc. is optimistic. We may want ~2-3 more digits there.
4. Once the dust settles and people are unblocked, I suggest presenting the overall structure during an engineering meeting. There's a lot to digest here, and everyone should understand how this code is structured.

I can look deeper when things have settled a bit.

tyleryasaka:
Closing in favor of the work being done on https://github.com/OriginProtocol/origin-js/tree/nick/upgradability

PR #292, opened 2018-07-31
Minor fractional usage listing updates

jordajm:
### Description:

This is a very small PR to tweak the fractional usage listings to stay in sync with work that is happening in the DApp. 

This should be able to be merged whenever it's convenient without breaking anything.

- remove support for "calendar step" because it got too complicated to support slot sizes other than daily / hourly at this point
- add support for `recurs` which the DApp will either be undefined or the string `"weekly"` to represent the weekly recurring events that the DApp now supports.


micahalcorn:
Is `recurs` supposed to be nested in `slots` or adjacent to it?

jordajm:
Doh! good catch @micahalcorn - I had an unnecessary reference to `recurs` - since it's nested in the slots object, it'll get returned in the array of slots.

PR #302, opened 2018-08-06
[WIP] Add search method to origin-js

franckc:

### Checklist:

### Description:

Add a search method to origin-js that calls the bridge server search API.

This is a work in progress (WIP):
- Missing unit tests
- I expect the implementation to change significantly once the contract upgradability work currently happening gets landed. I'm going to wait on that and fix the scalability issue regarding having to load all contracts before landing this PR...

Now, even though this is WIP, I'd appreciate any feedback on the current code especially since I'm totally new to JS ! Thanks :)

### DIFF
--- a/src/resources/listings.js
+++ b/src/resources/listings.js
@@ -83,7 +83,7 @@ class Listings extends ResourceBase {
         this.contractService.listingsRegistryContract
       )
     } catch (error) {
-      console.log('Contract not deployed', error)
+      console.log('Contract not deployed')
       throw error
     }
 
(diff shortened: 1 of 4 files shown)


micahalcorn:
Looking great 👍 Better than my JS 😁 I'm curious to see what the query argument will be once the DApp has filters and things beyond just a string value. I know that's not in v1, but you might consider going ahead with origin-js expecting an object or something assuming that the DApp isn't going to pass in a url-friendly query string.

src/resources/listings.js
```diff
@@ -99,6 +108,63 @@ class Listings extends ResourceBase {
     return range(0, Number(listingsLength))
   }
 
+  async search(query) {
+    // Issues an API call to the bridge server for searching listings.
+    // Args:
+    //   query(SearchQuery): query object.
+    // Returns:
+    //   list(int): listing ids that match the query.
+    if (query.rawQuery == '') {
```
tyleryasaka: It's generally considered best practice in javascript to use the strict equality operator `===`. See https://stackoverflow.com/a/359509

src/resources/listings.js
```diff
@@ -99,6 +108,63 @@ class Listings extends ResourceBase {
...
+      return []
+    }
+
+    // Make the search API call to the bridge server.
+    const url = encodeURI(
+      appendSlash(this.indexingServerUrl) + 'api/search/listings?query=' + query.rawQuery)
```
tyleryasaka: Just FYI javascript also has string interpolation, which I like. E.g. this could also be written

```javascript
appendSlash(`${this.indexingServerUrl}api/search/listings?query=${query.rawQuery}`)
```

src/resources/listings.js
```diff
@@ -99,6 +108,63 @@ class Listings extends ResourceBase {
...
+    const response = await this.fetch(url, { method: 'GET' })
+    if (response.status != 200) {
```
tyleryasaka: Same thing here with the strict inequality operator `!==`

src/resources/listings.js
```diff
@@ -99,6 +108,63 @@ class Listings extends ResourceBase {
...
+    const listingIds = await this.allIds()
+
+    // Get the listing registry contract deployed on the blockchain.
+    const contract = this.contractService.listingsRegistryContract
+    const listingsRegistry = await this.contractService.deployed(contract)
+
+    // Build a map of address -> id.
+    const addressToId = {}
+    await Promise.all(
+      listingIds.map(async id => {
```
tyleryasaka: This seems odd to me. The `map` method is used to return an array of mapped values. If you just want to iterate without returning anything, `forEach` is all you need.

That said I much prefer to use `map` and not actually modify things in place. I think there's a way to do this here with some refactoring. I'll come back to this.

test/resource_listings.test.js
```diff
@@ -165,6 +165,48 @@ describe('Listing Resource', function() {
...
+  describe('search', async () => {
+    it('should search listings', async () => {
+      const indexingServerUrl = 'http://hello.world'
+
+      const listings = new Listings({
+        contractService,
+        ipfsService,
+        undefined,
+        indexingServerUrl: indexingServerUrl
+      })
```
tyleryasaka: So this notation is shorthand for passing in a sort of implementation of named parameters. Actually, it's just a destructured javascript object.

This is equivalent to:
```
new Listings({
	contractService: contractService,
	ipfsService: ipfsService,
	undefined: undefined,
	indexingServerUrl: indexingServerUrl
})
```

So in other words you're actually passing in an object property named "undefined" - I'm pretty sure that's not what you intended. If you want to leave a specific named parameter out, you can just omit it from the call.

Another nit is that you can shorten the syntax for `indexingServerUrl: indexingServerUrl`. When you leave out the value of a key/value pair, the value is assumed to be a variable with the same name as the key.

To sum things up, this can all just be written as:

```
const listings = new Listings({
	contractService,
	ipfsService,
	indexingServerUrl
})
```

franckc:
Super useful feedback for me to learn JS @tyleryasaka - really appreciate it !
I'm going to update this CL based on marketplace changes....

franckc:
Domen implemented this.

PR #317, opened 2018-08-07
Source map type change

tomlinton:

### Checklist:

### Description:

- Changed the source map type to something that should provide a better experience when setting breakpoints in Chrome. It should also be faster to generate.
- Removed some docker stuff that is no longer used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,10 +8,6 @@
   "main": "./dist/index.js",
   "scripts": {
     "build": "node scripts/build.js",
-    "bridge:bash": "docker exec -it origin-bridge-container /bin/bash",
-    "bridge:build": "docker build ./bridge-container -t origin-bridge-image",
-    "bridge:start": "docker run -d -p 4000:4000 -p 5000:5000 -p 5432:5432 -p 6379:6379 -p 9200:9200 --name origin-bridge-container origin-bridge-image",
-    "bridge:stop": "docker stop origin-bridge-container && docker rm origin-bridge-container",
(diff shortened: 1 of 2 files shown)


micahalcorn:
Thanks @tomlinton 🙏 Were you able to test this and get a different debugging experience? I am running this and https://github.com/OriginProtocol/origin-dapp/tree/tomlinton/webpack-config but I still can't set breakpoints in most places using Chrome and Firefox.

tomlinton:
@micahalcorn with `inline-source-map` I had problems where I'd set a breakpoint in Chrome and it'd immediately unset. This addresses that for me. Did you experience that?

I still can't set breakpoints on every line though. Hope I didn't get your hopes up! I think because of how webpack transpiles things the original source code doesn't have a one to one line mapping with what is actually being executed in the browser. This means that not every line of the original source is breakpointable. I found I could always get a breakpoint when I needed it though, because there should be some line that is reasonably close. I didn't test it too extensively though.

There are other kinds of source maps we could try here. I ran across a handy graphic of the breakpoint support for each source map type, but I can't seem to find it again.

micahalcorn:
@tomlinton in my experience it's not so much that the breakpoint unsets but that it skips down to whatever line is mapped. I haven't noticed any improvement on that front, but I'm not consistently trying to set breakpoints in the exact same places.

Yeah, I had tried `eval-source-map`, which looked like it might be the most debug-friendly, but it runs into an inability to use `require`. 😕

tomlinton:
That's a shame, back to the drawing board on this then. :cry: We could still merge this PR though as the new source map should be faster to generate.

PR #326, opened 2018-08-10
Initial token implementation

### LINKED ISSUE #284: Token contract upgradeability
Because our needs of the token contract and best practices for Ethereum smart contracts are evolving, our token code will need to evolve as well. This issue tracks that work.

### LINKED ISSUE #295: Initial token implementation
The initial token implementation will include:

- Basic ERC-20 functionality
- Mint (create new) tokens
- Burn tokens
- Pause & unpause transfer of tokens

Additionally, it will be upgradeable as required by #284.

This will not include wiring up the token to any user flows.

cuongdo:
### Checklist:

### Description:

MVP for Origin token w/ ERC-20, minting, burning, pause transfers

Sorry for the size of this PR. A few things will make reviewing easier:

* All of the code in `contracts/test/openzeppelin-token` was ported from OpenZeppelin's ERC-20 tests with minimal (< 10 lines of code) changes. So, those don't require much review.
* The main token contract is `contracts/contracts/token/V001/OriginToken`.
* The token contract code is ported from OpenZeppelin's token contract code. I grouped functions based on the OpenZeppelin contract I based the code on.
* All token data is stored in `EternalStorage`, except `owner`. I didn't think it necessary to port that one state variable.
* I modeled the contract directory structure after the marketplace directory structure.
* When playing with contracts in Remix, it's useful to bump the gas limit to 7,000,000.
* There are more tests that need to be written. I'll do most of those in separate PRs.

I'm happy to walk through the code with any reviewers as people start freeing up from the marketplace & indexer work.

I might add more tests to this PR, but I'll save further features for another PR.

Resolves #284
Resolves #295

cc @nick @tyleryasaka @DanielVF @matthewliu


contracts/contracts/eternalstorage/EternalStorage.sol
```diff
@@ -0,0 +1,260 @@
+pragma solidity 0.4.23;
```
tyleryasaka: This contract has a lot going on. Did you make this from scratch or does this exist somewhere else?

contracts/contracts/mocks/EternalStorageMock.sol
```diff
@@ -0,0 +1,16 @@
+pragma solidity ^0.4.23;
```
tyleryasaka: I can understand the usefulness of this in remix. But I might be a little nervous if we had tests relying on the mock rather than the actual contract. Since each test only has to be written once, I think it would probably be worth it to use the actual contract in the tests. Especially considering what's at stake.

contracts/contracts/token/V001/BurnableToken.sol
```diff
@@ -0,0 +1,50 @@
+pragma solidity ^0.4.23;
+
+import "./StandardToken.sol";
+
+contract V001_BurnableToken is V001_StandardToken {
+  //
+  // Ported from OpenZeppelin's BurnableToken to use EternalStorage
```
tyleryasaka: Hmm... I hate that we have to diverge from OpenZeppelin to use EternalStorage. We lose the confidence we would otherwise inherit from OpenZeppelin. Really opens things up to potential bugs. Not that we can't do our own audits, but I doubt we will be able to get anywhere near as many eyes on our contracts as the OpenZeppelin contracts have.

cuongdo:
I understand your nervousness @tyleryasaka, but the problem is that we do need upgradability. 

It's either EternalStorage or using a delegate proxy. Because proxies delegate calls to another contract but use the storage of the proxy contract, it's vital that the implementation change its storage layout.

The OpenZeppelin AdminUpgradeabilityProxy code is evolving independently of their token code. Case in point:

https://github.com/OpenZeppelin/openzeppelin-solidity/commit/ef347ffccc1fd50b7f24b9c2d744eceef716ce16

I can't look at this diff and guarantee that the storage layout hasn't change from the prior rev.

The OpenZeppelin tests give some confidence that the behavior hasn't changed.

cuongdo:
The [prototype I sent earlier](https://github.com/cuongdo/origin-token-prototype/tree/master/contracts/eternal_prototype) has a more basic version of EternalStorage. It was based off this code:

https://github.com/rocket-pool/rocketpool/blob/master/contracts/RocketStorage.sol

contracts/contracts/mocks/EternalStorageMock.sol
cuongdo: This isn't used in tests at all. It's only for Remix.

cuongdo: I just removed one unneeded import of this file, which should make the lack of dependence on this clearer.

cuongdo:
Closing in favor of using proven token code and migrating balances between old and new contracts

PR #334, opened 2018-08-14
Handle missing block reference

### LINKED ISSUE #333: Failure to get block timestamp after recording a transaction
After creating a new listing, the DApp reported an error fetching the block.

screen shot 2018-08-13 at 8 49 48 pm

Wisdom from @micahalcorn:

> This is only used for the time reference in the blockchain confirmations dropdown. And we aren't supposed to treat a block time as gospel anyway. So I suggest just checking for the block, and if it doesn't exist, we use the current time.

micahalcorn:
Resolves #333 ⏲ 

- [x] Tests pass

### DIFF
--- a/src/resources/listings.js
+++ b/src/resources/listings.js
@@ -300,13 +300,6 @@ class Listings extends ResourceBase {
       console.error(error)
       throw new Error(`ETH Failure: ${error}`)
     }
-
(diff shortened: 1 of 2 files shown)


franckc:
Thanks for fixing so quickly !

PR #335, opened 2018-08-14
Apollo GraphQL server skeleton

franckc:

### Checklist:

### Description:
Implementation of the skeleton for a GraphQL server based on the Apollo framework.
This is a starting point. There is lot of remaining work to make it ready for production, including:
- Refine schema (should include offers, purchases, users 
- Add unit tests
- Add README with instructions on how to start the server
- Configure origin-box to start server
- and more...


daemon/apollo/index.js
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+// Type definitions define the "shape" of the data and specify
+// which ways the data can be fetched from the GraphQL server.
+const typeDefs = gql`
+  type Listing {
+    id: String
+    name: String
+    description: String
+    price: Float
```
sparrowDom: At some point we will probably add `ListingType` and `ListingCategory` here

DanielVF:
It visually looks good. What do I need to get data in to run it though?

Using the origin box DB with `{ Listings { name } }` gives me an error: "database "graphql" does not exist"

Using the origin box Elastic search with `{ Listings(query:"lake"){ name } }`  gives me error: "[index_not_found_exception] no such index, with { resource.type="index_or_alias" & resource.id="origin" & index_uuid="_na_" & index="origin" }"

DanielVF:
At least on my computer, the two files in `lib` can't find and use the node_modules inside `apollo`, since that's not in their directory or in a parent's directory. Is this just me, or do you have a node_modules in `daemon` that they are pulling from?

daemon/apollo/index.js
franckc: Yes, agreed the schema needs a *lot* more thinking. This is just a prototype.

I have started putting some thoughts into it but I don't have yet something ready to share.
At a high level I think it may look like this:
 - A core listing object including things like: Type, category, description, currency+price, pics, location, fractional data
- Custom fields that partner can use to store whatever data specific to their listings/apps. Would be garbage in/out.

franckc:
@DanielVF In order to run, it needs
- Postgres and Elasticsearch. Currently I have them running locally on my host, then I run the listener in the origin-box container after changing the configs in db.js and search.js to point to my host's IP. But really the better way is to update origin-box.
- I created the Postgres schema manually. Here is the schema I have for the listing table (this is only for prototyping - needs more thinking before productionizing it):

`
graphql-# \d listing
            Table "public.listing"
 Column |          Type           | Modifiers 
--------+-------------------------+-----------
 id     | character varying(32)   | not null
 data   | character varying(4096) | 
Indexes:
    "listings_pkey" PRIMARY KEY, btree (id)
`
Schema creation needs to be automated via origin-box setup. For bridge server we were using "flask db" to manage schemas so we'll need to look at something equivalent for Node...

franckc:
Thanks for pointing the lib issue. I'm clearly not proficient at npm packages :)

franckc:
@DanielVF @sparrowDom I pushed an update to this branch - pls let me know any new comments:
1. Changed directory structure. Now /listener, /apollo, /lib are under a common parent directory daemon/indexing which is where the npm modules are.
2. Fixed DB and Elasticsearch config to work with origin-box
3. Added support for DB migration.
4. Added README

daemon/indexing/migrations/20180815061230-add-listing-table.js
```diff
@@ -0,0 +1,53 @@
...
+  * This enables us to not have to rely on NODE_PATH.
+  */
+exports.setup = function(options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+  Promise = options.Promise;
+};
+
+exports.up = function(db) {
```
sparrowDom: This could be refactored to something like this: 
```
exports.up = function(db) {
  await executeSql(db, '20180815061230-add-listing-table-up.sql')
};

exports.down = function(db) {
  await executeSql(db, '20180815061230-add-listing-table-down.sql')
};

async executeSql(db, sqlFilePath) {
  var filePath = path.join(__dirname, 'sqls', sqlFilePath);
  return new Promise( function( resolve, reject ) {
    fs.readFile(filePath, {encoding: 'utf-8'}, function(err,data){
      if (err) return reject(err);
      console.log('received data: ' + data);

      resolve(data);
    });
  })
  .then(function(data) {
    return db.runSql(data);
  });
};
}
```

sparrowDom: Perhaps even moving executeSql function out of migration file into some sort of utils might make sense

DanielVF:
@franckc I made two small tweaks to the DB side.

I'm thinking we should use one of the postgres json datatypes to hold the `data` field, rather than a varchar. This will allow to later actually query and index off of the data inside the json string, as well as remove the current length limit.

https://www.postgresql.org/docs/9.5/static/datatype-json.html

daemon/indexing/migrations/20180815061230-add-listing-table.js
franckc: @sparrowDom This code is actually auto-generated by the db-migrate tool :)
See https://db-migrate.readthedocs.io/en/latest/Getting%20Started/installation/

franckc:
@DanielVF Thanks for fixing all this ! Yes, using the postgres json datatypes is a good idea ! I'm going to land this as is and let's iterate on subsequent PRs....

daemon/indexing/migrations/20180815061230-add-listing-table.js
sparrowDom: haha ok, didn't know that :)

PR #340, opened 2018-08-15
Move and rephrase origin-box suggestion

wanderingstan:


wanderingstan:
Holding off for now...

PR #347, opened 2018-08-16
GraphQLschema

franckc:

### Checklist:

- [] Update any relevant READMEs and [docs](https://github.com/OriginProtocol/docs)

### Description:
Proposal for an initial GraphQL schema.

Couple notes:
1.  This is a starting point - the main goal is to define the high level structure for objects and queries. Once this is in place, we can further flesh out the details such as the exact list of fields for each object.
2. I looked at some large GraphQL schemas like GitHub that are more "graph" oriented and have the concepts of "Nodes" and "Edges". I think it makes sense for those because it is a much more complex schema with relationships between dozens of different entities. In our case, our schema is pretty simple so I'm suggesting we keep it simple and flat. I don't think we need to add the extra complexity of making it more graph oriented.


daemon/indexing/apollo/index.js
```diff
@@ -4,37 +4,170 @@ var search = require('../lib/search.js')
...
+    buyer: User!
+    # Per contract definition: 1: Created, 2: Accepted, 3: Disputed
+    # NOTE: There is no "Finalized" status stored on-chain but could be useful
+    #       to compute it when resolving this object.
+    status: Int!
+  }
+
+  type Price {
+    currency: String!
+    amount: Float!
```
DanielVF: Prices should definitely not be floats. Javascript doesn't have enough precision to handle ETH.

We might have to have this be a string. (Which can then be parsed client-side by the big number library.)

daemon/indexing/apollo/index.js
```diff
@@ -4,37 +4,170 @@ var search = require('../lib/search.js')
...
+  }
+
+  # TODO: Add a status indicating if Listing is sold out.
   type Listing {
-    id: String
-    name: String
+    id: ID!
+    ipfsHash: ID!
+    seller: User!
+    title: String!
```
DanielVF: Some kind of category should be a part of even our draft work here.

daemon/indexing/apollo/index.js
```diff
@@ -4,37 +4,170 @@ var search = require('../lib/search.js')
...
+    filter: ListingFilter
+    orderBy: ListingOrderBy
+    page: Page
   }
 
   # The "Query" type is the root of all GraphQL queries.
   type Query {
-    Listings(query: String): [Listing],
+    Listings(query: ListingQuery!): [Listing],
+    Listing(id: ID!): Listing,
```
DanielVF: We need to allow querying down from Users and Offers as well.

A common use case would be to get a list of all my active offers. (Forgive the half thought out code below.)

```
Users(id:"me"){
  offers(active:True){
      id,
      ...,
      listing{
        id,
        name
      }
  }
}
```

DanielVF: If we are skipping the nodes and edges stuff, we still need a way to return the number of results for a paginated listing.

For example, if someone has 1,000 listings for sale, we need to be able to get listings between 100 and 150 listings, while also showing the total number of listings so that we can show the appropriate navigation options on the client.

daemon/indexing/apollo/index.js
franckc: I actually think ETH would be ok. AFAIK, Javascript uses 64 bits floats encoded with IEEE-754 which should give us something like 15 digits. So I can't really think of any item price in ETH that would cause precision loss ? But who knows, in the future we may support currencies with high inflation rate like zimbabwean dollar ! ;) Or possibly this API may get used by a language or library that deserializes JSON float using 32 bits. So overall better to use string.

And yes, also agreed that for doing math on those prices origin-js and the dapp should use a proper library like big-number or maybe currency.js

DanielVF: One ETH is 10^18 wei, and the contracts all operate on wei. This means that origin-js needs to be accurate-to-the-wei, and with it the search server.

daemon/indexing/apollo/index.js
franckc: Added.

daemon/indexing/apollo/index.js
franckc: Ah yep, makes sense. Changed to String.

franckc:
Based on review feedback, I added pagination and also a new User query.
The pagination is offset based rather than cursor based. Again, main reasoning is to keep things super simple to start with. And anyway, I doubt we'll need pagination for V1 since I expect there will be a very small number of listings and offers in the systems (I'd estimate a few thousands at most).

@DanielVF and @sparrowDom Let me know what you think.... Thanks :)

franckc:
Here is an example illustrating what a Listing query would look like:
screen shot 2018-08-19 at 12 13 00 am

franckc:
@sparrowDom @DanielVF How does this look in your opinion ? Ok if I land this as an initial version and we can iterate ? I'd like to get it in soon so that @sparrowDom can do the initial integration of origin-js with it :)

daemon/indexing/apollo/index.js
```diff
@@ -4,37 +4,268 @@ var search = require('../lib/search.js')
...
 // which ways the data can be fetched from the GraphQL server.
 const typeDefs = gql`
+  ######################
+  #
+  # Query output schema.
+  #
+  ######################
+
+  # When querying a set of items, the output is a page.
+  interface OutputPage {
```
sparrowDom: I now the pagination is not planned to be implemented yet, but since we have the interface laid out I would alter it a bit. When displaying paginated elements the DAPP will need to be aware of the current page & the maximum number of pages. For that reason I would change the interface to: 
```
interface OutputPage {
  pageNumber: Int # Current page number
  lastPageNumber: Int # The number of the last page
  }
```

daemon/indexing/apollo/index.js
```diff
@@ -4,37 +4,268 @@ var search = require('../lib/search.js')
...
+  #
+  # ORDER
+  #
+  enum OfferOrderField {
+    CREATION_DATE
+    STATUS
+  }
+
+  enum ReviewOrderField {
+    CREATION_DATE
```
sparrowDom: Maybe also sort by RATING_ASC and RATING_DESC

daemon/indexing/apollo/index.js
```diff
@@ -4,37 +4,268 @@ var search = require('../lib/search.js')
...
+  #  - Add location based filtering.
+  #  - Add fractional usage (e.g. availability) filtering.
+  input ListingFilter {
+    priceMin: inPrice
+    priceMax: inPrice
+    category: String
+    subCategory: String
+    locale: String
+    sellerAddress: String
+    buyerAddress: String
```
sparrowDom: Just a thought (not sure if it makes sense) to add `has_reviews`, `has_offers`

daemon/indexing/apollo/index.js
```diff
@@ -4,37 +4,268 @@ var search = require('../lib/search.js')
...
   }
 
   # The "Query" type is the root of all GraphQL queries.
   type Query {
-    Listings(query: String): [Listing],
+    Listings(searchQuery: String, page: Page, order: ListingOrder, filter: ListingFilter): ListingPage,
```
sparrowDom: We could add page size parameter here, that is limited by some maximum value. (something similar to what we have on the bridge-server at the moment: https://github.com/OriginProtocol/origin-bridge/blob/master/app/app_config.py#L42 )

daemon/indexing/apollo/index.js
franckc: Sounds good. Changed to the following to be consistent with the Page type used as input:
    num: Int!      # Current page number
    size: Int!     # Size of pages.
    total: Int!    # Total number of pages available.

daemon/indexing/apollo/index.js
franckc: Added.

daemon/indexing/apollo/index.js
franckc: If needed the caller can ask for the offers and reviews field. So not sure we need this ?
Leaving it out for now and east to add it later in case it turns out to be useful.

daemon/indexing/apollo/index.js
franckc: I think MaxResultsPerPage will be a constant used in the resolver code for Listings. In case the caller asks for page.size that is greater than MaxResultsPerPage, then the page size will get capped. Added a constant and a comment to reflect this.

PR #350, opened 2018-08-17
Token balance migration contract

cuongdo:

### Checklist:

### Description:

This is a prototype (with no comments or JS tests) of a token migration contract that:

1. Pauses the old token contract
2. Pauses the new token contract
3. Transfers ownership of the new token contract to the migration contract
4. For each balance in the old contract, mints the equivalent balance in the new contract. Minting can be done safely while paused, because it's restricted to the token contract owner. Remember: during this phase, the migration contract owns the new token contract, so only the migration contract can mint new tokens for now.
5. Finishes the migration, which transfers the new token contract to the rightful owner and prevents further migrations

This requires no changes to the `OriginToken` contract.

The second contract is for use with Remix, if you want to play with things. It's how I did a quick test of the concept. 

I'll write tests and improve the contract-level safeguards once @DanielVF and @nick give their thoughts on the approach.


cuongdo:
@DanielVF @nick What do you think of the overall approach? I tried to make this as simple as possible, and it's pretty agnostic to the token used as long as it can be paused and mint new tokens.

contracts/contracts/token/TokenMigration.sol
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  constructor(OriginToken _fromToken, OriginToken _toToken) public {
+    owner = msg.sender;
+    fromToken = _fromToken;
+    toToken = _toToken;
+  }
+
+  function migrateAccounts(address[] _holders) public onlyOwner notFinished {
+    for (uint i = 0; i < _holders.length; i++) {
```
DanielVF: Doing this in one for loop from solidity transaction probably isn't going work because of gas limits. If we have 20,000 token holders this will gas limit out far before that point.

I think the usual practice is to store a migrated-so-far index in the contract, and do X amount of migrations each transaction, updating the migrated-so-far index at the end of each mini-loop. Then just call the thing until it's done.

cuongdo: My plan is to call `migrateAccounts` for batches of accounts from a JS command-line tool through `web3`. Some JS needs to be involved, because the list of `address`es will come from retrieving past logs, which are only accessible to the JS. The JS would then perform checkpointing to ensure that it can restart upon crash or interruption.

I'm not sure how I could store an index, unless I stored all accounts to migrate as an array in the contract. I'm not clear on what the benefit to doing that is though.

DanielVF: I didn't pay enough attention when reading the code. Yeah, I think your way will work.  I like the token totals check being an integrated part of the finalize step.

cuongdo:
Thanks for the review @DanielVF. Merged minus the demo code, with some comments. JS wrapper and tests to come.

PR #356, opened 2018-08-21
Documentation fix

sparrowDom:
### Description:

Elasticsearch module is not present in the root `package.json` file. Thats is why before running the listener the current directory needs to be changed to `/app/daemon/indexing `.

### DIFF
--- a/daemon/indexing/README.md
+++ b/daemon/indexing/README.md
@@ -24,7 +24,7 @@ If you want to index data in Elasticsearch, start the elasticsearch container.
 
 Start the listener in the the origin-js container. Use --elasticsearch and/or --db options to pick the indexer(s).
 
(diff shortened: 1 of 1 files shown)


daemon/indexing/README.md
```diff
@@ -24,7 +24,7 @@ If you want to index data in Elasticsearch, start the elasticsearch container.
 
 Start the listener in the the origin-js container. Use --elasticsearch and/or --db options to pick the indexer(s).
 
-    docker exec -ti origin-js node daemon/indexing/listener/listener.js --elasticsearch --db
+    docker exec -w /app/daemon/indexing origin-js node listener/listener.js --elasticsearch --db
```
franckc: Ah yes, good catch ! Could you also update the instructions on line 43 below for starting the graphql server.

sparrowDom: yup ofc just changed it

PR #357, opened 2018-08-21
Listener Error Handling

DanielVF:
If there is an error loading information about an origin.js object, then the listener will skip that event and continue to the next. Because of the design of the Origin Protocol, there is zero guarantees that the associated IPFS data for a resource will be valid, or even there at all. Anyone can put whatever they want there.

When an error is raised when outputting to specific output handler (webhook, db, etc), the listener will attempt retries with increasing delays, up to two minutes. These retries will block all further event processing until the event goes through. If a maximum number of retries on one event has failed, then listener will quit, allowing it to be restarted from outside.

When an error is raised when getting event or block number information, the same retry strategy as for output errors is tried (increasing delays).

Added a listener specific readme.


daemon/indexing/listener/listener.js
```diff
@@ -182,9 +174,40 @@ async function runBatch(opts, context) {
   return lastLogBlock
 }
 
+async function withRetrys(fn) {
+  let tryCount = 0
+  while (true) {
+    try {
+      await fn() // Do our action.
+      return // it worked!
```
franckc: should this return whatever the fn returned ? just for future flexibility.

daemon/indexing/listener/listener.js
```diff
@@ -182,9 +174,40 @@ async function runBatch(opts, context) {
...
+      waitTime = Math.floor(waitTime * (1.2 - Math.random() * 0.4))
+      // Max out at two minutes
+      waitTime = Math.min(waitTime, 2 * 60 * 1000)
+      console.log('ERROR', e)
+      console.log(`will retry in ${waitTime / 1000} seconds`)
+      tryCount += 1
+      await new Promise(resolve => setTimeout(resolve, waitTime))
+    }
+    if (tryCount > 10) {
+      console.log('Maximum number of retrys reached')
```
franckc: nit: maybe add something like "exiting" to the message log. To make it clear.

daemon/indexing/listener/listener.js
```diff
@@ -182,9 +174,40 @@ async function runBatch(opts, context) {
...
+      // Now it's up to our enviroment to restart us.
+      // Hopefuly with a clean start, things will work better
+      process.exit(1)
```
franckc: Would throwing an uncaught exception be a safer way to exit ?
I was reading: https://nodejs.org/api/process.html#process_process_exit_code

daemon/indexing/listener/listener.js
```diff
@@ -252,7 +305,7 @@ async function postToWebhook(urlString, json) {
   }
   return new Promise((resolve, reject) => {
     const req = http.request(postOptions, res => {
-      if(res.statusCode == 200){
```
franckc: nit: in general when writing our JS code, shouldn't we always favor using the === vs the == operator ? Same comment for line 294 below.

daemon/indexing/listener/listener.js
```diff
@@ -182,9 +174,40 @@ async function runBatch(opts, context) {
   return lastLogBlock
 }
 
+async function withRetrys(fn) {
```
franckc: nit: Add a comment explaining this retries up to N times, with exponential backoff.

daemon/indexing/listener/listener.js
```diff
@@ -182,9 +174,40 @@ async function runBatch(opts, context) {
...
+    try {
+      await fn() // Do our action.
+      return // it worked!
+    } catch (e) {
+      // Roughly double wait time each failure
+      let waitTime = Math.pow(100, 1 + tryCount / 6)
+      // Randomly jiggle wait time by 20% either way. No thundering herd.
+      waitTime = Math.floor(waitTime * (1.2 - Math.random() * 0.4))
+      // Max out at two minutes
+      waitTime = Math.min(waitTime, 2 * 60 * 1000)
```
franckc: nit: define maxRetryCount and maxWaitTime as constants so that if we decide to change them later on it's very easy to do so.

daemon/indexing/listener/listener.js
DanielVF: Yeah, that would be nice. Good thinking.

daemon/indexing/listener/listener.js
DanielVF: Just throwing here doesn't exit, since it's inside an async call. 

It feels like it should be okay, given that it will have been many minutes since we last wrote any data we care about.

DanielVF:
Added in the tweaks from your review. Should be ready to merge.

sparrowDom:
You guys are moving fast, sorry for missing this review

PR #358, opened 2018-08-21
Listener - Continue where it left off after a restart

DanielVF:
`--continue-file=path` - Will start following events at the block number defined in the file, and will keep this file updated as it listens to events.

This should be merged after PR #357 to keep the history simple.

### DIFF
--- a/daemon/indexing/listener/README.md
+++ b/daemon/indexing/listener/README.md
@@ -8,7 +8,7 @@ The listener will let you know one or more times about an event. Make sure your 
 
 To allow the listener to be compatible with [infura.io](https://infura.io/), it does not use subscriptions, only API queries.
 
-## Running
+# Running
 
 First you'll need a blockchain network to listen to. To get a local network work, you can start up the origin box, or you can run `npm start run` from the origin.js directory.
 
@@ -16,7 +16,9 @@ A simple way to see the listener in action:
 
     node scripts/listener.js --verbose
 
-### Command line options
+## Command line options
+
+Output:
 
 `--verbose` Output json for all event information to stdout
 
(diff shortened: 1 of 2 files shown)


daemon/indexing/listener/listener.js
```diff
@@ -136,6 +139,26 @@ async function liveTracking(config) {
   check()
 }
 
+function getLastBlock(config){
+  if(config.continueFile == undefined || !fs.existsSync(config.continueFile)){
+    return 0
```
franckc: If config file not found, I would print an error and exit rather than defaulting to block 0

daemon/indexing/listener/README.md
```diff
@@ -26,8 +28,12 @@ A simple way to see the listener in action:
 
 `--db` Experimental support for recording listings directly into postgres (see instructions for setting up the db [here](../README.md))
 
+Events:
+
+`--continue-file=path` Will start following events at the block number defined in the file, and will keep this file updated as it listens to events.
```
franckc: nit: mention it is JSON format and provide an example of a continue file.

daemon/indexing/listener/listener.js
```diff
@@ -136,6 +139,26 @@ async function liveTracking(config) {
...
+  }
+  const json = fs.readFileSync(config.continueFile, {encoding:"utf8"})
+  const data = JSON.parse(json)
+  if(data.lastLogBlock){
```
franckc: style nit: space before parenthesis ? it seems that is the style we use in origin-js repo ?

daemon/indexing/listener/listener.js
DanielVF: It's going to be very normal that there's no continue file the first time you run it. Also for now if you want to reindex, you can just delete the continue file.

PR #359, opened 2018-08-21
Allow listener to continue where it left off

DanielVF:
Github falsely believes PR #358 was merged.

This PR you are reading includes that old PR, plus the tweaks from the previous PR's review.

### DIFF
--- a/daemon/indexing/listener/README.md
+++ b/daemon/indexing/listener/README.md
@@ -8,7 +8,7 @@ The listener will let you know one or more times about an event. Make sure your 
 
 To allow the listener to be compatible with [infura.io](https://infura.io/), it does not use subscriptions, only API queries.
 
-## Running
+# Running
 
 First you'll need a blockchain network to listen to. To get a local network work, you can start up the origin box, or you can run `npm start run` from the origin.js directory.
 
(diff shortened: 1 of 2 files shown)


franckc:
This PR is for a merge into danielvf/listenererrorhandling - did you mean to merge to develop ?

DanielVF:
image

I figured out what happened with the merging. In the past when I've worked on multiple features faster than they are merged, I've started the next feature off the previous feature's branch (as in the illustration). Usually what happens is after the first is merged in, the base on the second PR then updates automatically to develop, and then is ready to be merged in. In this case though, the previous PR got merged with a squash commit? and so all those commits disappeared and this PR ended up having all the other PR's previous commits as well as it's own.

image

So I made a new local branch, cherry picked out the real commits into it, and force pushed it up under this name. Should be good to merge.

micahalcorn:
@DanielVF interesting case. 🙁 Yes, the decision was made to use "Squash and merge" [back in May](https://github.com/OriginProtocol/origin-js/issues/123#issuecomment-388571621). Do we need to revisit that?

franckc:
Yes, I confirm that I did a "squash and merge" on your previous PR.

micahalcorn:
Unless I'm mistaken, my learning on the squash situation is that if I branch from a feature branch, I need to ensure that my base doesn't get PR'd and merged before I get a chance to merge. Or I just need to hard rebase my branch after it all happens. This seems like an edge case that should not change our squashing strategy, yeah?

DanielVF:
Yeah, we don't have to change the squashing strategy, it just wasn't what I was expecting.

PR #367, opened 2018-08-23
Listener - works on mainnet

DanielVF:
- Configurable web3 and IPFS urls
- Catch up with by reading chunks of blocks at a time, rather than all at once
- Tweaks to getLogs call so it will work on Infura

(Note that it won't actually work on mainnet just by changing the urls, unless you have the contract addresses, which we aren't currently shipping in origin.js)

### DIFF
--- a/daemon/indexing/listener/listener.js
+++ b/daemon/indexing/listener/listener.js
@@ -11,26 +11,18 @@ const db = require('../lib//db.js')
 // ---------------
 
 // Todo
-// - Allow configuring web3 endpoint, network, and IPFS endpoint
-// - When catching up, work in smaller batches
 // - Handle blockchain splits/winners
-// - Possibly switch lastLogBlock to be closer to last found block
 // - Include current-as-of block numbers in POSTs
 // - Persist starting point in DB
-// - Perhaps send related data as it was at the time of the event, not crawl time
+// - Perhaps send related data as it was at the time of the event, not as of crawl time
 // - Possible configurable log levels
 
-const web3Provider = new Web3.providers.HttpProvider('http://localhost:8545')
-const web3 = new Web3(web3Provider)
-const o = new Origin({
-  ipfsDomain: 'origin-js',
-  ipfsGatewayProtocol: 'http',
-  ipfsGatewayPort: 8080,
-  web3
-})
+let web3
+let o
 
 MAX_RETRYS = 10
 MAX_RETRY_WAIT_MS = 2 * 60 * 1000
+MAX_BATCH_BLOCKS = 30000
 
 // -----------------------------
 // Section 1: Follow rules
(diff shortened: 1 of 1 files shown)


daemon/indexing/listener/listener.js
```diff
@@ -104,10 +96,26 @@ const LISTEN_RULES = {
 // Section 2: The following engine
 // -------------------------------
 
+function setupOriginJS(config){
+  const web3Provider = new Web3.providers.HttpProvider(config.web3Url)
```
franckc: nit: add a console.log statement to show which provider and ipfs url is being used. Will be useful when checking in production that the listener got started with proper config.

daemon/indexing/listener/listener.js
```diff
@@ -11,26 +11,18 @@ const db = require('../lib//db.js')
...
-  ipfsGatewayProtocol: 'http',
-  ipfsGatewayPort: 8080,
-  web3
-})
+let web3
+let o
 
 MAX_RETRYS = 10
 MAX_RETRY_WAIT_MS = 2 * 60 * 1000
+MAX_BATCH_BLOCKS = 30000
```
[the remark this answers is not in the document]
sparrowDom: Maybe it would make sense to put all this into `.env` file. As @franckc mentions depending on the volume of listings on our contracts we might need to tweak these numbers. At it is nice we can do that without the necessary code changes.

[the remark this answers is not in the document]
sparrowDom: Actually this might change between runs of the listener perhaps a better way would be to write the default values in .env file, an make it overridable with cmd parameters.

daemon/indexing/listener/listener.js
```diff
@@ -431,6 +438,10 @@ const config = {
   // Verbose mode, includes dumping events on the console.
   verbose: args['--verbose'],
   // File to use for picking which block number to restart from
-  continueFile: args['--continue-file']
+  continueFile: args['--continue-file'],
+  // web3 provider url
+  web3Url: args['--web3-url'] || 'http://localhost:8545',
+  // ipfs url
+  ipfsUrl: args['--ipfs-url'] || 'http://origin-js:8080',
```
sparrowDom: I think `ipfsUrl` and `web3Url` would be better located in the `.env` file instead of as command line parameters.

DanielVF: I think adding the ability to configure out of environment variables would be out of scope for this PR, but I'll add an issue for it.

daemon/indexing/listener/listener.js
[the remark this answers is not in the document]
DanielVF: I've dropped the max_batch_blocks down to 3000 (half a day) and added a note.

daemon/indexing/listener/listener.js
DanielVF: Added console.logs for which web3 and which IPFS url are used.

franckc:
@sparrowDom based on Daniel's comments, you ok with removing your blocking review ? then I'll merge this in...

PR #368, opened 2018-08-23
Include our contract addresses in origin.js

DanielVF:
Currently our non-local contract addresses only live in contracts/build json files and are not in version control and cannot be used outside of NPM releases.

This PR creates an addresses.json file that holds all our official contract addresses. These addresses are automatically injected into our contracts definitions.

Advantages:

- Clear at all times what our official contract addresses are
- Version control for our contract addresses
- You can work on the public networks from git versions of origin.js

Addresses used:

I've used the current contract addresses used in https://demo.originprotocol.com/

### DIFF
--- a/src/services/contract-service.js
+++ b/src/services/contract-service.js
@@ -5,9 +5,10 @@ import KeyHolderLibrary from './../../contracts/build/contracts/KeyHolderLibrary
 import UserRegistryContract from './../../contracts/build/contracts/UserRegistry.json'
 import OriginIdentityContract from './../../contracts/build/contracts/OriginIdentity.json'
 import OriginTokenContract from './../../contracts/build/contracts/OriginToken.json'
-
 import V00_MarketplaceContract from './../../contracts/build/contracts/V00_Marketplace.json'
 import V01_MarketplaceContract from './../../contracts/build/contracts/V01_Marketplace.json'
+
+import ContractAddresses from './../../contracts/releases/addresses.json'
 
 import bs58 from 'bs58'
 import Web3 from 'web3'
@@ -44,7 +45,8 @@ class ContractService {
         this[name].networks = Object.assign(
           {},
           this[name].networks,
-          options.contractAddresses[name]
+          ((ContractAddresses[name]|{}).network|{}),
+          (options.contractAddresses[name]|{})
         )
       } catch (e) {
         /* Ignore */
--- a/src/services/contract-service.js
+++ b/src/services/contract-service.js
@@ -44,9 +44,9 @@ class ContractService {
       try {
         this[name].networks = Object.assign(
           {},
-          ((ContractAddresses[name]|{}).network|{}), // start with hard coded addresses
+          ((ContractAddresses[name]||{}).network||{}), // start with hard coded addresses
           this[name].networks, // override with addresses in contracts/build
-          (options.contractAddresses[name]|{}) // override with passed in options
+          (options.contractAddresses[name]||{}) // override with passed in options
         )
       } catch (e) {
         /* Ignore */
--- a/src/services/contract-service.js
+++ b/src/services/contract-service.js
@@ -44,9 +44,9 @@ class ContractService {
       try {
         this[name].networks = Object.assign(
           {},
-          this[name].networks,
-          ((ContractAddresses[name]|{}).network|{}),
-          (options.contractAddresses[name]|{})
+          ((ContractAddresses[name]|{}).network|{}), // start with hard coded addresses
+          this[name].networks, // override with addresses in contracts/build
+          (options.contractAddresses[name]|{}) // override with passed in options
         )
       } catch (e) {
         /* Ignore */
--- a/contracts/releases/addresses.json
+++ b/contracts/releases/addresses.json
@@ -1,0 +1,26 @@
+{
+  "listingsRegistryContract": {
+    "networks": {
+      "3": { "address": "0x67057a6ddbc10d1c75a8812f770854f8627fc7b6" },
+      "4": { "address": "0x4f041f3ec338f85048831d1dff15433d3cb3983f" }
+    }
+  },
+  "listingsRegistryStorageContract": {
+    "networks": {
+      "3": { "address": "0x2293f5592885de85c68b03f69522023957ad5133" },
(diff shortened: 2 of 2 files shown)


wanderingstan:
I haven't run the code, but would this preclude users from using the dapp with other contract addresses on Rinkeby/Ropsten? 

E.g. I've testing our Rinkeby migrations by re-deploying our contracts to throwaway addresses and testing the data migration there.

nick:
I like having official contracts checked in to source control. From looking at the code the order of precedence is: 

1. Contract specified in options
2. Checked in contract addresses
3. Truffle build file

As @wanderingstan pointed out, perhaps we should switch the last two and let the local truffle build file take precedence?

wanderingstan:
Working on the data migration script has made me even more in favor of having contracts and their abi/contract json files in source control.

A few thoughts:

### Where to store contract info?

Rather than introduce a new json file format as in this PR, I'd propose we just use the truffle json files as-is. I.e. rather than zipping the `build` dir as I did last time, we could include it wholesale unzipped. Path would be something like `origin-js/contracts/releases/0.7.1/build` 

A proliferation of formats could get messy. One format means less work at release time (just use truffle's output), and all code anywhere that needs to read contract info can be standardized. (E.g. scripts like the listing migration script, and any future scripts like that.) 

### How to use deployed contract addresses by default?

Rather than "injecting" contract addresses in `contract-service.js` could we instead add logic in truffle's build script that notices and goes "Hey, there is no `contracts/build` directory! We'll copy over the build dir from the latest release to get them started." This way the contract addresses and ABI's in the build directory always remain the gold standard, unless overwritten by manually set options. 

Things could get hard to debug when there are too many places that could be changing what contracts origin-js is *actually* using.

tyleryasaka:
I agree that the Truffle build file should take precedence over the additional `addresses.json` file. I also share @wanderingstan 's concern that this might be getting to complex. My only question around checking in the Truffle build files themselves is whether we will lose the deployed contract addresses on test nets when we run the build and migration locally.

src/services/contract-service.js
```diff
@@ -43,8 +44,9 @@ class ContractService {
       try {
         this[name].networks = Object.assign(
           {},
-          this[name].networks,
-          options.contractAddresses[name]
+          ((ContractAddresses[name]|{}).network|{}), // start with hard coded addresses
```
tyleryasaka: Can you explain why you are using `|` here?

I see that it's the bitwise OR operator. But why is it being used? At a glance it seems like it might produce unintended behavior. For example, `undefined|{}` evaluates to `0`.

Even if it does behave as intended, I'm hesitant to use obscure operators. Maybe it's just me but I've never seen anyone do this in Javascript before.

Why not just use the regular OR (`||`)?

wanderingstan:
@tyleryasaka asks:

>  My only question around checking in the Truffle build files themselves is whether we will lose the deployed contract addresses on test nets when we run the build and migration locally.

Not to worry: When you run migrations locally, the new contract addresses are _merged_ into the build files. E.g. when you migrate to local test net, those contract addresses would be added alongside the pre-existing Rinkeby and Ropsten. You would lose the deployed addresses on test nets only when you do a migration specifically to those networks.

wanderingstan:
@DanielVF - How hard would it be to have this function take the contract addresses from truffle build files rather than adding `addresses.json`? I suspect it'd be a matter of (a) finding `build/contracts` dir from latest release (b) iterating over the contract files (e.g. `ListingsRegistry.json`) and finding the rinkeby/ropsten/mainnet addresses of that contract. 

This would mean we have only one contract json format, and be one step (aka something that could go wrong) when making a release. 

**Alternative:** That said, I still wonder if we should instead add an initial step to `[scripts/build.js](https://github.com/OriginProtocol/origin-js/blob/master/scripts/build.js)` to check if a `build` directory exists and —if not— copy over the one from latest build. This would mean that we're back to only two possibilities: addresses in `build` dir or addresses passed as options. (Experienced developers will know to look in `build` directory to find addresses. I'm worried about confusing developers who see the dapp working on a network that is nowhere to be found in `build` dir.)

src/services/contract-service.js
DanielVF: Thanks! Fixed!

DanielVF:
@wanderingstan This function is already using the addresses from the contracts/build directory. If you you wanted use that instead of an addresses.json, the build files would just need to be checked into source code control, and we can throw away this PR. Or am I misunderstanding your question?

wanderingstan:
@DanielVF sorry, I think (or thought?) we have two issues going on:

1. Getting build files (including addresses) in source control, i.e. an unzipped `build` dir. I think everyone agrees that's a good idea. Presumably stored in `origin-js/contracts/releases/X.X.X/build/`
2. Automatically using contracts from latest release when on test networks. <sub>(*When user hasn't done their own `truffle migrate` to those networks) </ssub>

My comment above was about how best to accomplish (2). I'm worried about adding a new `addresses.json` file that we need to remember to keep in sync with address in the release build files.

tomlinton:
I've been looking at this for the deployments. What I'm thinking is that the CI/CD pipeline should do the following (we already do most of this, the only real new bit is steps 3 and 4):

1. Detect a PR that changes the `contracts` directory
2. Run any tests
3. Deploy contracts to the relevant network e.g. for the development deployment environment we'll have our own network using ganache, for staging we'll use rinkeby. 
4. Add a commit to the branch with the truffle build artifacts
5. Mark the PR as mergeable

That way `origin-js` will always have up to date contract artifacts. I'll be able to do builds of `origin-js` for deployment that will always have the correct artifacts, and we can decouple things like the event listener and the faucet nicely. We then just add a script to parse the contract addresses like @wanderingstan  described for when webpack is being used with `origin-js`.

cuongdo:
Sorry for being late to the discussion. I've only recently started digging into Truffle build processes. I don't think we should add another file that must be manually synced, because we know how problematic this can be from having had to manually sync contract JSON files to `bridge-server` until recently.

There are some good ideas in this thread, and I want to explore them further. I'm also up for implementing whatever solution makes the most sense for us. I'll set up a meeting to discuss.

DanielVF:
Since we are now taking a different approach, I'm closing this PR.

PR #371, opened 2018-08-23
Token transfer sender/recipient whitelist

cuongdo:
### Checklist:

### Description:

For a limited time, this enforces a whitelist of allowed senders and
recipients for Origin token transfers. The intent is to allow a gradual
launch and comply with regulations.


cuongdo:
Currently, this only includes basic tests to demonstrate how the whitelist would work. We'd have real, JS tests if we agree on the high-level approach.

There is little actual logic. It's mainly safeguards and comments. All OpenZeppelin token tests still pass.

cc @joshfraser @matthewliu @franckc

franckc:
Read the code. Really nice work at implementing this in a very readable and simple way. I also like that it emits events so it is very transparent what addresses are getting added/removed to the whitelist. And having the option to postpone the "doomsday clock" is nice to :)

Overall I like this approach. It is very easy to reason about and I don't think it would cause an issue from an auditing or community perspective. +1 with moving forward in this direction.

joshfraser:
+1 

Thanks for the quick turnaround on this @cuongdo. Agreed that this looks like a clean and straightforward approach.

cuongdo:
Thanks for the reviews everyone. Merging without demo test code. Tests are tracked as part of #372.

PR #382, opened 2018-08-29
Arbitration refunds. Relative finalization window.

nick:

### Checklist:

### Description:

Adds support for refunds in arbitration as well as relative finalization windows.

### DIFF
--- a/contracts/contracts/marketplace/v00/Marketplace.sol
+++ b/contracts/contracts/marketplace/v00/Marketplace.sol
@@ -8,7 +8,7 @@ pragma solidity ^0.4.23;
  */
 
 contract IArbitrator {
-  function createDispute(uint listingID, uint offerID) external returns (uint);
+  function createDispute(uint listingID, uint offerID, uint refund) external returns (uint);
 }
 
 contract ERC20 {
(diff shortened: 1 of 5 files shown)


contracts/contracts/marketplace/v00/Marketplace.sol
```diff
@@ -184,6 +184,9 @@ contract V00_Marketplace {
     require(msg.sender == listing.seller);
     require(offer.status == 1); // Offer must be in state 'Created'
     require(listing.deposit >= offer.commission);
+    if (offer.finalizes < 1000000000) { // Relative finalization window
+      offer.finalizes = now + offer.finalizes;
```
tyleryasaka: Why can't we just do this in javascript?

nick: This line is basically deciding whether offer.finalizes should be treated as an absolute timestamp or an offset in seconds. If it's an offset (ie < 1,000,000,000) then it's added to the current time. We're basically letting offer.finalizes represent two things to save some gas and keep the contract code simple. We have to do it on-chain as otherwise the seller could set it to `now` and immediately finalize without necessarily delivering the goods.

tyleryasaka: I see now. Makes sense.

PR #383, opened 2018-08-29
Return a `unitsAvailable` property for each listing resource

### LINKED ISSUE #379: Handle Listing Quantity
There are a few cases that deal with a listing's quantity, which we should prioritize in the following order:

1. A single-unit listing sells - We need to be able to determine that it is "sold out" and prevent more sales. (Q3)
2. A sale occurs for a multi-unit listing - We need to know how many available units are remaining and conditionally prevent more sales. (not likely Q3)
3. A buyer purchases multiple units - We need to limit the quantity purchased to a number less than equal to the available supply. (not Q3)

Since the initial quantity is no longer stored in the contract, I suppose that quantity value(s) will now be in IPFS. We need to decide which values can be calculated on demand by the DApp (if indexed data can be quickly retrieved) and which values need to be stored or should be denormalized. There may also be some discussion about whether or not values should be automatically [de/in]cremented or if they should be adjusted by the seller.

@franckc @tyleryasaka

tyleryasaka:
### Checklist:

### Description:

The `unitsAvailable` property is derived from the `units` property on the listing ifps data and the number of offers created for a listing.

Does not handle offer cancellation/rejection, but will need to handle that once those pieces are implemented.

Uses a default `units` value of `1` if not present in the ipfs data.

Resolves #379


micahalcorn:
Thx @tyleryasaka 🙏

PR #386, opened 2018-08-30
Core listing schema

franckc:
### Checklist:

### Description:
Proposal for the "core" listing schema to use for our V1 launch.
This is based on discussions we had in this [doc](https://docs.google.com/document/d/12xtHzYRsxsa-V6hMAP-aC8xl3nzfm1pdYcQ3wFQ9kwE/edit) and during this [meeting](https://docs.google.com/document/d/10xtZ1sZE9Bbz6Goe4RG8iJbBghpX2jOEHoVIK6fBpww/edit).

Couple notes:
 - This PR is only to agree on the listing schema. It does *not* include wiring this up with origin-js code (that will be handled as a separate PR).
  - For V1, the only class of listing we'll support is "unitListing". But I added "fractionalListing" to illustrate how the schema would play out in future.
  - For V1, we won't support "Custom" schema that can be used to store specific properties associated with a (category, subCategory) as proposed in the doc. But it can be added later.
 - The definition of the "Location" object is included to illustrate what it would look like. but it won't be used in V1.
 - This schema would replace the current "listing.json", "fractional-listing.json" and "unit-listing.json" schemas.

Here is an [example of a listing data](https://gist.github.com/franckc/2c741ad71fdd0d04b9720289e46d5ee0) that validates against this schema.

### DIFF
--- a/src/schemas/listing-core.json
+++ b/src/schemas/listing-core.json
@@ -1,0 +1,243 @@
+{
+  "$id": "http://schema.originprotocol.com/listing-core-v1.0.0",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Origin Protocol core listing schema",
+  "definitions": {
+    "moneyAmount": {
+      "type": "object",
+      "title": "An amount of money in a specified currency",
+      "properties": {
+        "currency": {
+          "type": "string"
+        },
+        "amount": {
+          "type": "string"
+        }
+      },
+      "required": [
+        "currency",
+        "amount"
+      ]
+    },
+    "location": {
+      "type": "object",
+      "properties": {
+        "street": {
+          "type": "string"
+        },
+        "extendedStreet": {
+          "type": "string"
+        },
+        "locality": {
+          "type": "string"
+        },
+        "region": {
+          "type": "string"
+        },
+        "box": {
+          "type": "string"
+        },
+        "postalCode": {
+          "type": "string"
+        },
+        "country": {
+          "type": "string"
+        },
+        "coord": {
+          "type": "object",
+          "properties": {
+            "latitude": {
+              "type": "integer",
+              "minimum": -90,
+              "maximum": 90
+            },
+            "longitude": {
+              "type": "integer",
+              "minimum": -180,
+              "maximum": 180
+            }
+          },
+          "required": [
+            "latitude",
+            "longitude"
+          ]
+        }
+      },
+      "required": [
+        "street",
+        "locality",
+        "region",
+        "country"
+      ]
+    },
+    "media": {
+      "type": "object",
+      "title": "Representation of a media asset such as an image or a video",
+      "properties": {
+        "url": {
+          "type": "string",
+          "format": "uri"
+        },
+        "contentType": {
+          "type": "string"
+        },
+        "description": {
+          "type": "string"
+        },
+        "dimension": {
+          "type": "object",
+          "properties": {
+            "height": {
+              "type": "integer"
+            },
+            "width": {
+              "type": "integer"
+            }
+          },
+          "required": [
+            "height",
+            "width"
+          ]
+        }
+      }
+    },
+    "timeSlot": {
+      "type": "object",
+      "required": [
+        "startDate",
+        "endDate",
+        "isAvailable"
+      ],
+      "properties": {
+        "startDate": {
+          "type": "string",
+          "format": "date-time"
+        },
+        "endDate": {
+          "type": "string",
+          "format": "date-time"
+        },
+        "isAvailable": {
+          "type": "boolean"
+        },
+        "price": {
+          "$ref": "#/definitions/moneyAmount"
+        }
+      },
+      "required": [
+        "startDate",
+        "endDate",
+        "isAvailable",
+        "price"
+      ]
+    },
+    "unitListing": {
+      "type": "object",
+      "properties": {
+        "unitsAvailable": {
+          "type": "integer"
+        },
+        "price": {
+          "$ref": "#/definitions/moneyAmount"
+        }
+      },
+      "required": [
+        "unitsAvailable",
+        "price"
+      ]
+    },
+    "fractionalListing": {
+      "type": "object",
+      "properties": {
+        "timeIncrement": {
+          "type": "string"
+        },
+        "calendarStep": {
+          "type": "string"
+        },
+        "slots": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/timeSlot"
+          }
+        }
+      }
+    },
+    "listingClass": {
+      "type": "object",
+      "title": "Support for Unit and Fractional listing types",
+      "properties": {
+        "name": {
+          "type": "string",
+          "enum": [
+            "unit",
+            "fractional"
+          ]
+        },
+        "fields": {
+          "anyOf": [
+            {
+              "$ref": "#/definitions/unitListing"
+            },
+            {
+              "$ref": "#/definitions/fractionalListing"
+            }
+          ]
+        }
+      },
+      "required": [
+        "name",
+        "fields"
+      ]
+    }
+  },
+  "type": "object",
+  "properties": {
+    "schemaVersion": {
+      "type": "string"
+    },
+    "class": {
+      "$ref": "#/definitions/listingClass"
+    },
+    "category": {
+      "type": "string"
+    },
+    "subCategory": {
+      "type": "string"
+    },
+    "locale": {
+      "type": "string"
+    },
+    "title": {
+      "type": "string"
+    },
+    "description": {
+      "type": "string"
+    },
+    "expiry": {
+      "type": "string",
+      "format": "date-time"
+    },
+    "media": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/media"
+      }
+    },
+    "commission": {
+      "$ref": "#/definitions/moneyAmount"
+    },
+    "securityDeposit": {
+      "$ref": "#/definitions/moneyAmount"
+    }
+  },
+  "required": [
+    "schemaVersion",
+    "class",
+    "category",
+    "subCategory",
+    "locale",
+    "title",
+    "description"
+  ]
+}
--- a/src/schemas/listing-core.json
+++ b/src/schemas/listing-core.json
@@ -1,9 +1,9 @@
 {
-  "$id": "http://schema.originprotocol.com/listing-core-v1.0.0",
   "$schema": "http://json-schema.org/draft-06/schema#",
+  "$id": "http://schema.originprotocol.com/listing-core-v1.0.0",
   "title": "Origin Protocol core listing schema",
   "definitions": {
-    "moneyAmount": {
+    "money": {
       "type": "object",
       "title": "An amount of money in a specified currency",
       "properties": {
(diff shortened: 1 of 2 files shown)


franckc:
@sparrowDom @DanielVF FYI since this is relevant for search.

src/schemas/listing-core.json
```diff
@@ -0,0 +1,243 @@
...
+      "$ref": "#/definitions/listingClass"
+    },
+    "category": {
+      "type": "string"
+    },
+    "subCategory": {
+      "type": "string"
+    },
+    "locale": {
+      "type": "string"
```
wanderingstan: Instead of `locale`, I'd prefer `language` or `lang`, to match js and html conventions, respectively. While there is some debate on meanings of "local" vs "language"*, I feel we should just match javascript and go with `language`. 

*e.g. See "4. Locale versus Natural Language " section of https://www.w3.org/TR/ltli/

src/schemas/listing-core.json
```diff
@@ -0,0 +1,243 @@
...
+            }
+          },
+          "required": [
+            "height",
+            "width"
+          ]
+        }
+      }
+    },
+    "timeSlot": {
```
wanderingstan: @tyleryasaka @jordajm - In our fractional implementation, I recall that the availability data was stored in a _different_ IPFS blob from the main listing, given its editability. Is that right? 

@franckc , if that's right, I don't think it makes sense to put fractional data in listing-core.

src/schemas/listing-core.json
franckc: Sounds good. Changed to _language_.

src/schemas/listing-core.json
```diff
@@ -0,0 +1,243 @@
+{
+  "$id": "http://schema.originprotocol.com/listing-core-v1.0.0",
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "title": "Origin Protocol core listing schema",
+  "definitions": {
+    "moneyAmount": {
+      "type": "object",
+      "title": "An amount of money in a specified currency",
```
wanderingstan: I'm not totally familiar with the guts of json schema. Will `title` ever be shown to users? If so, it should be replaces with a generic identifier that can be localized, e.g. https://github.com/OriginProtocol/origin-dapp/blob/master/public/schemas/housing.json#L15

src/schemas/listing-core.json
```diff
@@ -0,0 +1,243 @@
...
+      ]
+    },
+    "unitListing": {
+      "type": "object",
+      "properties": {
+        "unitsAvailable": {
+          "type": "integer"
+        },
+        "price": {
+          "$ref": "#/definitions/moneyAmount"
```
wanderingstan: @micahalcorn , we are [still using](https://github.com/OriginProtocol/origin-dapp/blob/59609cce764ad48d2f5a9e8769717c902dec428b/src/components/listing-create.js#L5) `react-json-schema` for rendering our forms, and I recall that it didn't handle anything too fancy like these json pointers. 

I get worried introducing such fanciness at the 11th hour... 😬 Your thoughts?

src/schemas/listing-core.json
franckc: No, title does not get shown to users. It simply provides a short description of the purpose of the object. See https://json-schema.org/understanding-json-schema/reference/generic.html

src/schemas/listing-core.json
micahalcorn: I actually have not worked with `react-jsonschema-form` at all. But it looks like [there is some support](https://github.com/mozilla-services/react-jsonschema-form#schema-definitions-and-references).

franckc: react-json-schema does support references to local definitions. See https://github.com/mozilla-services/react-jsonschema-form#schema-definitions-and-references

But it does *not* support anyOf - see https://github.com/mozilla-services/react-jsonschema-form#json-schema-supporting-status

Now, per previous comment, I was thinking the core listing schema would be used by origin-js  to do validation but DApp can use a different schema for rendering the listing creation form.

src/schemas/listing-core.json
```diff
@@ -0,0 +1,243 @@
...
+      },
+      "required": [
+        "currency",
+        "amount"
+      ]
+    },
+    "location": {
+      "type": "object",
+      "properties": {
+        "street": {
```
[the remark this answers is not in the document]
franckc: Yes, that would go into the "extendedStreet" field.

src/schemas/listing-core.json
```diff
@@ -0,0 +1,243 @@
...
+          "type": "string"
+        },
+        "extendedStreet": {
+          "type": "string"
+        },
+        "locality": {
+          "type": "string"
+        },
+        "region": {
```
tomlinton: We might need more fields here? It is hard to fit all kinds of international addressing systems into one schema. We might need more than just region. Google deals with this by using
`administrative_area_level_1`,... `administrative_area_level_5`. E.g. in the US:

- `administrative_area_level_1` corresponds to a state
- `administrative_area_level_2` corresponds to a county

But in other countries, they can just be adapted to whatever administrative divisions there are. I don't know too much about other addressing systems, but in places like Japan there might be quite a few administrative divisions, e.g. prefectures, districts, cities, wards.

src/schemas/listing-core.json
wanderingstan: Ahhh, understood. I didn't realize this was for _validation_. Though should have clicked that we were in the `origin-js` repo! (For others reading, here is [json schema validation spec](https://datatracker.ietf.org/doc/draft-handrews-json-schema-validation/))

> ...DApp could keep using those schemas to get the data from the user at listing creation time, then data gets passed to origin-js, which creates a JSON object from it, ...

Is the idea that origin-js would _transform_ the data coming from our old schemas do fit into this schema?

src/schemas/listing-core.json
wanderingstan: @franckc are we following (aka copying) some other standard on location?

Meta comment: I'm still a little perplexed by our strategy of not wanting to attempt doing listing schemas until we can "do them right" (and fear of getting locked into a bad standard), but then OTOH now including a location standard—something notoriously hard to "do right"!

src/schemas/listing-core.json
wanderingstan: Gotcha. Did not realize this was for validation. And good to learn about `react-jsonschema-form` being able to handle local definitions.

src/schemas/listing-core.json
franckc: Good feedback @tomlinton and @wanderingstan This was primarily modeled after schema.org. Location is out of scope for V1 anyway so let me just take it out of this schema. We need more research/thinking - I agree better to not put it our there if we are not completely sure it is the right schema... :)

franckc:
Thanks everyone for the feedback - super useful ! I made some updates to incorporate your comments and I also updated the [example listing](https://gist.github.com/franckc/2c741ad71fdd0d04b9720289e46d5ee0) to illustrate what the data would look like.

src/schemas/listing-core.json
tyleryasaka: Yeah we use `ajv` for the origin-js schemas, and it does handle refs. I was able to get it to work before.

src/schemas/listing-core.json
```diff
@@ -0,0 +1,184 @@
...
+        }
+      ]
+    },
+    "category": {
+      "type": "string"
+    },
+    "subCategory": {
+      "type": "string"
+    },
+    "language": {
```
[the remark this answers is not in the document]
franckc: @sparrowDom Yes, I think this is a good idea in the long-term. Though I'm worried about the amount of changes required on the DApp side to support this. Let's do this incrementally - I'll get the current schema in and then we can iterate on adding multi-languages.

src/schemas/listing-core.json
```diff
@@ -0,0 +1,243 @@
...
+        }
+      },
+      "required": [
+        "street",
+        "locality",
+        "region",
+        "country"
+      ]
+    },
+    "media": {
```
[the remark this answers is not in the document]
franckc: Updated

src/schemas/listing-core.json
```diff
@@ -86,15 +85,15 @@
     "unitListing": {
       "type": "object",
       "properties": {
-        "unitsAvailable": {
+        "unitsForSale": {
```
micahalcorn: Why this? Not everything is "for sale", per se. 🤔

franckc: There was code added recently to compute in origin-js "unitsAvailable" based on "unitsForSale" and "unitsSold". See https://github.com/OriginProtocol/origin-js/blob/8c1b19abd33fad514dd09ad6a6145ec52ce6818c/src/resources/marketplace.js#L88

So I just sticked to that convention... Happy to change it if it is cleaner.

src/schemas/listing-core.json
```diff
@@ -0,0 +1,243 @@
...
+        "coord": {
+          "type": "object",
+          "properties": {
+            "latitude": {
+              "type": "integer",
+              "minimum": -90,
+              "maximum": 90
+            },
+            "longitude": {
+              "type": "integer",
```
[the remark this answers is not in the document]
franckc: Good catch. We decided localization schema is out of scope for V1 but I updated the doc so that we remember when we implement it.

src/schemas/listing-core.json
```diff
@@ -0,0 +1,243 @@
...
+          "type": "string"
+        },
+        "country": {
+          "type": "string"
+        },
+        "coord": {
+          "type": "object",
+          "properties": {
+            "latitude": {
+              "type": "integer",
```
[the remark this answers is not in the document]
franckc: Good catch. We decided localization schema is out of scope for V1 but I updated the doc so that we remember when we implement it.

src/schemas/listing-core.json
```diff
@@ -0,0 +1,243 @@
...
+        "box": {
+          "type": "string"
+        },
+        "postalCode": {
+          "type": "string"
+        },
+        "country": {
+          "type": "string"
+        },
+        "coord": {
```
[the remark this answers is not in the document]
franckc: Good catch. We decided localization schema is out of scope for V1 but I updated the doc so that we remember when we implement it.

src/schemas/listing-core.json
micahalcorn: Oh, right. I forgot about that distinction. So "available" is more like "remaining" and "forSale" is more like the total supply? 😕

franckc: Yes, you are correct. I'm pretty bad at naming things though so happy to make changes...
Do you think this would be more clear ?
 - unitsTotal (field in the schema): total number of units available
 - unitsRemaining (computed by origin-js. difference of unitsTotal-unitsSold): number of units remaining
 - unitsSold (computed by origin-js based on offer data)
Though to your earlier point everything is not for sale so "unitsSold" is not great.... hmmm any better idea ?

micahalcorn: I think that is definitely an improvement. 👍

micahalcorn:
@nick mentioned that we should potentially mirror the on-chain values in IPFS and validate that they match. We may want to maintain that information for when listings are withdrawn. I think there were potentially other benefits to mirroring, but I don't remember what they were.

franckc:
@micahalcorn @nick What "on-chain values" are you specifically thinking we should mirror in IPFS ?
For a listing, there is currently a very few pieces of data we store on-chain. See [contract](https://github.com/OriginProtocol/origin-js/blob/master/contracts/contracts/marketplace/v01/Marketplace.sol#L37).
Are you suggesting we should store arbitrator and deposit in IPFS ?

Note that this PR deliberately only addresses Listing data to keep the scope manageable - Offer has lot more data but is out of scope for this PR...

nick:
I think we should mirror everything on-chain in IPFS too. Doesn't have to be in this pull-request, but I think it'd be valuable going forward for looking at the evolution of the data in transaction history and also for arbitration. Of course there's nothing to stop saying one thing in IPFS and another on-chain... but we can potentially penalize that kind of bad behavior in the future.

joshfraser:
Don't fully understand the specifics here, but in general, I prefer to have a single source of truth rather than 2 copies of the same data that could get out of sync. There's a risk that developers would put too much trust in the ipfs versions. If you have to constantly check that the two copies match, you may as well check the blockchain every time, no?

nick:
Yea the blockchain should be the source of truth but the data can change over time and is deleted once a listing is withdrawn or an offer is finalized or withdrawn in order to get a gas refund. After the data is deleted, it'll only be retrievable (via regular web3 calls) from either IPFS or the event log.

We can make all this transparent to the developer in origin-js so they shouldn't have to worry about it. `offer.price` should read the blockchain if the offer is still active... or IPFS if not.

src/schemas/listing-core.json
franckc: Ok. I'm changing things to unitsTotal, unitsRemaining, unitsSold

PR #394, opened 2018-08-31
📙 Remove all but ipfs urls from listing pictures before indexing.

DanielVF:
Instead of removing all pictures data, let's keep IPFS urls.

### DIFF
--- a/daemon/indexing/listener/listener.js
+++ b/daemon/indexing/listener/listener.js
@@ -281,9 +281,9 @@ async function handleLog(log, rule, contractVersion, context) {
   const userAddress = log.decoded.party
   const ipfsHash = log.decoded.ipfsHash
 
-  //TODO: remove binary data from pictures in a proper way.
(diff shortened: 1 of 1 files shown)


franckc:
Some merge conflicts due to recent schema changes.
Also, there has been some recent updates to the DataStore so I don't expect any listing to have pictures as Data URLs anymore. See this [code](https://github.com/OriginProtocol/origin-js/blob/master/src/services/_data-store-adapter.js#L84).

So I'm not sure we need to do any filtering anymore ? Except if it is to guard against listing data uploaded directly by partners/malicious actors and that bypassed origin-js code.
But maybe for that we should add some additional guards in marketplace.getListing to detect those potentially malformed/malicious listings ?

DanielVF:
I'll close this PR, no need for it anymore.

We do need to make sure we always keep it a throwing error in origin.js when loading a listing using http/https images, since anyone could track people by using them.

PR #395, opened 2018-08-31
📙 GraphQL convention is to use camelCase connection names in the queryroot

DanielVF:
GraphQL convention is to use camelCase connection names in the query root

(see Facebook, GitHub, and Shopify APIs)

https://developer.github.com/v4/query/
https://help.shopify.com/en/api/graphql-admin-api/reference/queryroot

This is a breaking change, and we’ll need to update the dapp search branch later.


franckc:
@sparrowDom FYI

PR #396, opened 2018-08-31
📙 Discovery - Switch listing connection to use Elasticsearch

DanielVF:
I think it makes sense for the discovery server to just pull from Elasticsearch, rather than half the methods use postgres and half elasticsearch. This should reduce the amount of code we have to write, make the discovery server more reliable, and simplify the storage side.


franckc:
I think in the future there will be query that are actually simpler to run against DB rather than ES. So part of me would be inclined to use ES only for full-text search and the DB for everything else. But let's see what uses cases come up and then we can decide...

PR #397, opened 2018-08-31
📙 Discovery - Use Elasticsearch for searching all listings

DanielVF:
Use Elasticsearch for all listings searches, rather than using DB for all listings and elasticsearch for text searching listings.


DanielVF:
Made style fix.

PR #399, opened 2018-09-01
📙 Discovery - switch connections to using “nodes”

DanielVF:
I like github/facebook's schema's style better, using just "nodes" for the actual object list. I think it's easier to read when compared with aways duplicating the name of the thing we are accessing.

    listings{nodes{id}} <-- new
      vs.
    listings{listings{id}}  <-- old

Plus it's a little more standard.

@sparrowDom (This change will affect you)


sparrowDom:
Cool, @DanielVF thanks for the heads up!

PR #402, opened 2018-09-04
 📙 Discovery - Offer Basics

DanielVF:
This brings the basics of offer support. I've also commented out non-implemented parts of the schema, so that generated docs match what should work.

Offers:

- Offers are indexed
- Search offers from root.
- See a listing’s offers.
- See an offer’s listing.
- Search for a buyer’s offers.

Other:

- Commented out placeholder schema elements and resolvers
- Make marketplace.getOffer return text statuses not integers status numbers. (This follows the way offers’ status in marketplace.getListing are done.)

### DIFF
--- a/daemon/indexing/apollo/index.js
+++ b/daemon/indexing/apollo/index.js
@@ -19,11 +19,11 @@ const typeDefs = gql`
   ######################
 
   # When querying a set of items, the output is a page.
-  interface OutputPage {
-    num: Int!      # Current page number
-    size: Int!     # Size of pages.
-    total: Int!    # Total number of pages available.
-  }
+  #interface OutputPage {
+    # num: Int!      # Current page number
+    # size: Int!     # Size of pages.
+    # total: Int!    # Total number of pages available.
+  #}
 
   type Price {
     currency: String!
@@ -32,49 +32,48 @@ const typeDefs = gql`
 
   type User {
     walletAddress: ID!   # Ethereum wallet address
-    identityAddress: ID  # ERC 725 identity address.
-    listings(page: Page, order: ListingOrder, filter: ListingFilter): ListingPage
-    offers(page: Page, order: OfferOrder, filter: OfferFilter): OfferPage
-    reviews(page: Page, order: ReviewOrder, filter: ReviewFilter): ReviewPage
+    # identityAddress: ID  # ERC 725 identity address.
+    # listings(page: Page, order: ListingOrder, filter: ListingFilter): ListingPage
+    # offers(page: Page, order: OfferOrder, filter: OfferFilter): OfferConnection
+    # reviews(page: Page, order: ReviewOrder, filter: ReviewFilter): ReviewPage
   }
 
   enum OfferStatus {
-    CREATED
-    ACCEPTED
-    DISPUTED
-    # Note: There is no "Finalized" status stored on the Offer contract on-chain.
-    #       This is computed by the indexing server.
-    FINALIZED
+    created
+    accepted
+    finalized
+    buyerReviewed
+    disputed
   }
 
   type Offer {
     id: ID!
     ipfsHash: ID!
-    listingId: ID!
     buyer: User!
+    seller: User!
     status: OfferStatus!
+    affiliate: ID,
+    price: Price! 
+    listing: Listing!
   }
 
-  type OfferPage implements OutputPage {
-    num: Int!
-    size: Int!
-    total: Int!
-    offers: [Offer]
+  type OfferConnection {
+    nodes: [Offer]!
   }
 
-  type Review {
-    ipfsHash: ID!
-    reviewer: User!
-    text: String!
-    rating: Int!
-  }
+  # type Review {
+  #   ipfsHash: ID!
+  #   reviewer: User!
+  #   text: String!
+  #   rating: Int!
+  # }
 
-  type ReviewPage implements OutputPage {
-    num: Int!
-    size: Int!
-    total: Int!
-    reviews: [Review]
-  }
+  # type ReviewPage implements OutputPage {
+  #   num: Int!
+  #   size: Int!
+  #   total: Int!
+  #   reviews: [Review]
+  # }
 
   # TODO: Add a status indicating if Listing is sold out.
   type Listing {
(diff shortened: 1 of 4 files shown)


daemon/indexing/lib/search.js
```diff
@@ -12,11 +12,13 @@ var client = new elasticsearch.Client({
...
-const indexName = 'origin'
-
-// Name of the index type used for storing listings data.
-const listingsType = 'listing'
+// Elasticsearch index and type names for our data
+// Elasticsearch is depreciating storing different types in the same index.
+// (and forbids it unless you enable a special flag)
+const LISTINGS_INDEX = 'listings'
+const LISTINGS_TYPE = 'listing'
+const OFFER_INDEX = 'offers'
```
franckc: Note that when making the image for the origin-box elasticsearch container, we currently create the "origin" index: https://github.com/OriginProtocol/origin-box/blob/master/dockerfiles/elasticsearch#L11

So we'll need to update that to create "listings" and "offers".

daemon/indexing/lib/search.js
```diff
@@ -113,16 +115,74 @@ class Listing {
...
+        priceEth: hit._source.priceEth,
       }
       listings.push(listing)
     })
     return listings
   }
 }
 
+class Offer {
+  /*
```
franckc: nit: JSdoc comments actually start with "/**" - my bad for using "/*" when I created this file...
Also it is @param vs @params... again my bad :(

daemon/indexing/lib/search.js
```diff
@@ -113,16 +115,74 @@ class Listing {
...
+   * Indexes an Offerr
+   * @params {object} offer - JSON offer data from origin.js
+   * @throws Throws an error if indexing operation failed. 
+   */
+  static async index(offer, listing){
+    const resp = await client.index({
+      index: OFFER_INDEX,
+      type: OFFER_TYPE,
+      id: offer.id,
```
franckc: Isn't offer id scoped to a listing ?
See https://github.com/OriginProtocol/origin-js/blob/master/contracts/contracts/marketplace/v01/Marketplace.sol#L161
Perhaps in order to generate a unique id for indexing purposes we could concatenate listing.id + offer.id ?

daemon/indexing/lib/search.js
```diff
@@ -113,16 +115,74 @@ class Listing {
...
+        priceEth: offer.priceEth,
+        status: offer.status
+      }
+    })
+  }
+
+  static async get(id) {
+    const resp = await client.get({id: id, index: OFFER_INDEX, type: OFFER_TYPE})
+    if(!resp.found){
+      throw "Offer not found"
```
franckc: nit: I'm a JS noob so asking :) Should we prefer throwing an Error rather than a String as a general convention in our stack ?

src/adapters/v00/marketplace_adapter.js
```diff
@@ -1,3 +1,5 @@
+const OFFER_STATUS = ['error','created', 'accepted', 'finalized', 'buyerReviewed']
```
franckc: Changes in this file must also be applied to src/adapters/v01/marketplace_adapter.js

src/adapters/v00/marketplace_adapter.js
```diff
@@ -265,6 +267,8 @@ class MarkeplaceAdapter {
       if (e.event === 'OfferData') {
         rawOffer.status = '4'
       }
+      // Translate status number to string
+      rawOffer.status = OFFER_STATUS[rawOffer.status]
```
franckc: Will this require updates to the DApp ?
If yes, let's make sure to land both this PR and the companion DApp PR at about the same time so that we don't break the DApp for developers for too long...

src/adapters/v00/marketplace_adapter.js
franckc: I think this is missing status "disputed" which is 3.
See https://github.com/OriginProtocol/origin-js/blob/master/contracts/contracts/marketplace/v01/Marketplace.sol#L55

And actually it looks like the contracts and this file are out of sync. In the code below, 'OfferFinalized' event should set rawOffer.status to '4' and 'OfferData' should set it to 5.

Out of scope for this PR but we should really write some unit tests to catch these type of things :)

daemon/indexing/lib/search.js
DanielVF: Elasticsearch creates the indexes automatically when you first put in a new document into it - it  doesn't seem to require them created ahead of time.

daemon/indexing/lib/search.js
DanielVF: *nods* Yes, this is the full, unique offer id, like "999-001-557-2"

daemon/indexing/lib/search.js
DanielVF: Will update.

daemon/indexing/lib/search.js
DanielVF: Will update

src/adapters/v00/marketplace_adapter.js
DanielVF: Yes, there appear to be three different ways the status comes back from three different methods in the adapter. :(

DanielVF:
Okay @franckc, this should be ready for a review again.

daemon/indexing/apollo/index.js
```diff
@@ -231,51 +243,63 @@ const resolvers = {
...
-      return {}
+    price(offer) {
+      return {currency: 'ETH', amount: offer.priceEth}
     },
-    reviews(user, args) {
-      // TODO: load reviews for the user, handle pagination, filters, order.
-      return {}
+    listing(offer, args, context, info) {
+      const requestedSubFields = info.fieldNodes[0].selectionSet.selections 
+      const isIdOnly = requestedSubFields.filter(x=>x.name.value !== "id").length == 0
```
franckc: I'm probably annoying you with my nits... :(
use === 0 rather than == 0

daemon/indexing/lib/search.js
franckc: To be more specific, what I'm a bit worried about is using dynamic mapping (which we would if we don't create the index) vs explicit mapping. See some [doc](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping.html) explaining the difference.

I think in the long-term I'd be more comfortable if we would specify an explicit mapping so that we don't run into issues when partners send us garbage docs. Let's discuss...

daemon/indexing/lib/search.js
franckc: Perfect.

src/adapters/v00/marketplace_adapter.js
```diff
@@ -1,3 +1,5 @@
+const OFFER_STATUS = ['error','created', 'accepted', 'disputed', 'finalized', 'buyerReviewed']
```
franckc: Could you also make those changes in v01/marketplace_adapter.js ?

DanielVF: There's only one adapter file, both versions share it.

PR #403, opened 2018-09-04
Implement a Listing abstraction

franckc:
### Checklist:

### Description:
This PR implements a "Listing" object that can be used as a stable interface exposed by origin-js to the DApps for accessing listing data.

For context, I realized I needed something like this to be able to roll out the new listing schema and after talking with @tyleryasaka and @micahalcorn it turned out it was something they were already thinking would be good to have in place.

Notes:
 - The new core listing schema is used to store data in IPFS.
 - The concept of "schema adapters" allows to load listing stored using old schemas. This makes it possible to support backward compatibility. There is no support for forward compatibility - it seems reasonable that a DApp developer would have to upgrade her origin-js version in order to interact with listings created with a new version of the schema.
 - The code that reads and writes listing data to IPFS is encapsulated in a ListingIpfsStore class. This should make it easy to swap a different type of storage if needed in the future (ex: filecoin ?)
  - The Listing object directly inherits from a variety of properties that are read from the blockchain (ex: offers, events, status, etc...). These properties are likely to change in future versions so ideally they should get abstracted. To keep the scope of this PR manageable, let's have discussions and handle this as a separate PR.

There is a number of changes that will be required on the DApp side to use this new Listing interface, including:
  - Some renames to use the fields 'title', 'category', 'subCategory'
  - Updates for using the Money object for price, commission and securityDeposit


DanielVF:
I like it.

Schema looks nice.

I guess throwing is the right way to go for validation errors, since those would really be a programatic error, rather than a user error at this level.

franckc:
@nick As FYI since he wrote a lot of the marketplace code.

src/resources/listing.js
```diff
@@ -0,0 +1,225 @@
...
+      subCategory:    data.subCategory,
+      language:       data.language,
+      title:          data.title,
+      description:    data.description,
+      media:          data.media,
+      expiry:         data.expiry,
+    }
+
+    // Unit data.
+    if (listing.type === 'unit') {
```
nick: Is it worth treating unit and fractional listing types as fundamentally different now that they're the same on-chain?

franckc: They are identical onchain but differ off-chain in the IPFS data. This method handles deserializing the data from off-chain, that's why unit vs fractional are handled differently.

Now, you are right that for our V1 release we won't support fractional. But I thought would be good to have it here as a placeholder. Since fractional usage schema is purely additive, I was thinking that when we finalize it after our V1 release, we can just put it in schema V1. We won't have to put in a schema V2, which will save us from writing a new SchemaAdapterV2.

Having said that, if this is too confusing, I can definitely remove any reference to fractional for now.

src/resources/listing.js
```diff
@@ -0,0 +1,226 @@
...
+const ajv = new Ajv({allErrors: true})
+// To use the draft-06 JSON schema, we need to explicitly add it to ajv.
+ajv.addMetaSchema(require('ajv/lib/refs/json-schema-draft-06.json'))
+ajv.addSchema(schemaV1)
+
+
+//
+// Listing is the main interface exposed by Origin Protocol to access data from a listing.
+//
+class Listing {
```
tyleryasaka: I think this looks great, but with our current setup resources are publicly exposed. I actually think we do need to have a listing resource so that people can do things like `origin.listings.create`, etc. But that resource should contain the public methods.

If something is not public facing, we've been making it a `service`. I think making this a `listing-service` might make sense here, unless you have any other suggestions.

src/resources/listing.js
tyleryasaka: Yeah fractional listings differ enough in ipfs data format that I think it makes sense to make them separate types. Fractional listings have the concept of time slots and the restriction that multiple bookings for the same item should not overlap.

test/resource_listing.test.js
```diff
@@ -0,0 +1,148 @@
...
+chai.use(chaiAsPromised)
+
+const expect = chai.expect
+
+import { Listing, ListingIpfsStore} from '../src/resources/listing.js'
+import sinon from 'sinon'
+import goodListing from './data/listing-valid'
+
+
+describe('Listing', () => {
```
tyleryasaka: 🎉 hooray for testing

src/resources/listing.js
tyleryasaka: One other thought: `service` doesn't necessarily make sense here to me either. This seems like a different type of thing than our contract service, for example.

What about `model`? We could start introducing the concept of models into this codebase. So this would be under `src/models/listing-model.js`

PR #419, opened 2018-09-06
fix compiler version

tyleryasaka:

### Checklist:

### Description:

Our solc version specified in package.json has been out of sync with our contracts. This went undetected because our package-lock.json file has not been properly synced for some time.

- Sync package-lock
- Upgrade contracts to solc ^0.4.24
- Fix resulting warnings & issues

### DIFF
--- a/contracts/contracts/Migrations.sol
+++ b/contracts/contracts/Migrations.sol
@@ -1,4 +1,4 @@
-pragma solidity ^0.4.23;
+pragma solidity ^0.4.24;
 
 contract Migrations {
   address public owner;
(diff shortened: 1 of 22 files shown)


micahalcorn:
@tyleryasaka I get a bunch of compilation errors after this merge and @crazybuster says it's because I need a newer version of truffle. Is it possible that I'm getting away with my existing [`4.1.8`](https://github.com/OriginProtocol/origin-js/blob/master/package.json#L77) during `install:dev` but that a clean repo would install something later that is, in fact, necessary?

PR #421, opened 2018-09-06
Sparrow dom/search

sparrowDom:
### Checklist:

### Description:
- Add the search interface to `origin-js`
- Add generic search filters to GraphQl schema
- Change pagination description in GraphQl

**Dependant PR:** https://github.com/OriginProtocol/origin-dapp/pull/396


daemon/indexing/apollo/index.js
```diff
@@ -167,23 +167,33 @@ const typeDefs = gql`
...
+    name: String!
+    value: String!
+    valueType: ValueType!
+    operator: FilterOperator!
   }
 
   # The "Query" type is the root of all GraphQL queries.
   type Query {
-    listings(searchQuery: String, page: Page, order: ListingOrder, filter: ListingFilter): ListingPage,
+    listings(searchQuery: String, page: Page, order: ListingOrder, filters: [ListingFilter!]): ListingPage,
```
franckc: What does the syntax [ListingFilter!] mean exactly ? The filters argument should be optional, so shouldn't this just [ListingFilter] ?

src/index.js
```diff
@@ -9,6 +9,8 @@ import store from 'store'
 
 const defaultBridgeServer = 'https://bridge.originprotocol.com'
 const defaultIpfsDomain = 'gateway.originprotocol.com'
+const defaultApolloServer = 'http://apollo.originprotocol.com'
```
franckc: Let's use discovery (more generic) rather than Apollo (specific to framework used for implementation):
`const defaultDiscoveryServer = 'https://discovery.originprotocol.com'`

src/resources/marketplace.js
```diff
@@ -15,18 +15,23 @@ const storeKeys = {
 }
 
 class Marketplace extends Adaptable {
+
   constructor({
     contractService,
     ipfsService,
-    fetch,
     indexingServerUrl,
+    apolloServer,
```
franckc: Same comment, let's use discovery.

src/resources/marketplace.js
```diff
@@ -15,18 +15,23 @@ const storeKeys = {
 }
 
 class Marketplace extends Adaptable {
+
   constructor({
     contractService,
     ipfsService,
-    fetch,
     indexingServerUrl,
```
franckc: Let's get rid of indexingServer in the origin-js codebase - it is not used by anything AFAICT.

src/resources/marketplace.js
```diff
@@ -108,6 +113,46 @@ class Marketplace extends Adaptable {
     })
   }
 
+  async search(searchQuery, listingType, filters = []) {
+
```
franckc: style nit: remove blank lines.

src/resources/marketplace.js
```diff
@@ -108,6 +113,46 @@ class Marketplace extends Adaptable {
     })
   }
 
+  async search(searchQuery, listingType, filters = []) {
```
franckc: style nit: since this is going to be a publicly exposed method used by DApp developers, let's document it well. I have been trying to encourage folks to use [JSDoc](http://usejsdoc.org/about-getting-started.html) format (but if anyone else has a better suggestion, please lmk).

src/resources/marketplace.js
```diff
@@ -108,6 +113,46 @@ class Marketplace extends Adaptable {
...
+          }
+          `
+        }).join(',')}]
+      ) {
+        pageNumber
+        itemsPerPage
+        totalNumberOfPages
+        nodes {
+          id
+
```
franckc: nit: remove blank line

src/resources/marketplace.js
```diff
@@ -108,6 +113,46 @@ class Marketplace extends Adaptable {
...
+        }
+      }
+    }`
+
+    return this.fetch(`${this.apolloServer}:${this.apolloServerPort}`, {
```
franckc: Do you need to call .json() on the response to parse the response ? Something like:
```
const resp = this.fetch(...)
return resp.json()
```

daemon/indexing/apollo/index.js
sparrowDom: The syntax means that the filters argument is optional and that it must not contain null values. It can be an empty list, but not a list where any value is `null`.

src/index.js
sparrowDom: good point

src/resources/marketplace.js
sparrowDom: Http request might fail with different sort of errors. Like server not reachable, server returned 500 error, JSON was malformed.... And for each of those errors we would need to catch them separately in `origin-js` and re-throw them, so that client (dapp in this case) can show the correct error message. 

My logic was just return the promise, and let client handle different sorts of errors anyway it wishes. 

Can you maybe think of a way to better handle this?

DanielVF:
It feels awkward to have the discovery server querying tied into the marketplace code. Since it's entirely generic, it feels more like it should live as a services/DiscoveryService?

We can talk later today about it.

daemon/indexing/apollo/index.js
franckc: For what reason would you want to pass an empty filters list ?

franckc:
@sparrowDom Is this ready for final round of review ?

sparrowDom:
@franckc @DanielVF Yup agreed. Discovery should be a separate resource separated from marketplace. Doing the changes now

sparrowDom:
@DanielVF @franckc this PR is ready for the final review.

src/index.js
```diff
@@ -10,7 +11,8 @@ import store from 'store'
 
 const defaultBridgeServer = 'https://bridge.originprotocol.com'
 const defaultIpfsDomain = 'gateway.originprotocol.com'
-const defaultDiscoveryServer = 'https://discovery.originprotocol.com'
+//const defaultDiscoveryServer = 'https://discovery.originprotocol.com'
```
franckc: Let's follow the same pattern as other configs. Default is what will be used in prod.
Then for local development we use env files:
https://github.com/OriginProtocol/origin-box/blob/master/envfiles/origin-dapp.env
https://github.com/OriginProtocol/origin-dapp/blob/master/dev.env

franckc:
Let's merge this puppy and then we can iterate on improvements.

PR #428, opened 2018-09-08
Allow sellers to initiate disputes

micahalcorn:
### Checklist:

### Description:

This allows the seller to initiate a dispute during the `Accepted` stage as described [here](https://github.com/OriginProtocol/origin-js/issues/365#issuecomment-418532841).


micahalcorn:
@cuongdo sorry, I didn't think about this. Are we actually planning to deploy/ship both v00 and v01 if they are identical?

nick:
I think the plan is to only deploy v01. The reason there is v00 and v01 right now is so we can demonstrate and verify upgradability... ie that deploying a new contract won't have any unintended side effects in the dapp.

@micahalcorn can you make the same change to v01?

micahalcorn:
Sure thing 👍 Should I revert the change to v00 or leave it?

nick:
I think we agreed to only make changes in v01 going forward, so revert v00 yea

PR #431, opened 2018-09-10
Make origin-js configurable from .env file

sparrowDom:
### Checklist:

### Description:
- making properties in `origin-js` configurable from `.env` file
- it takes `dev.env` file into consideration regarding what are all the properties that should be defined in `.env` file. Any if any is missing throws an error
- once the origin.js server/client files are built in the `dist` folder configurations from `.env` file are copied into those files and not ready from environmental variables

#### Post Merge Steps
- Resolves Issue: https://github.com/OriginProtocol/origin-js/issues/437
- Also merge this branch once this is merged: https://github.com/OriginProtocol/origin-dapp/pull/480#pullrequestreview-154153057

### DIFF
--- a/dev.env
+++ b/dev.env
@@ -1,0 +1,7 @@
+BRIDGE_SERVER=https://bridge.originprotocol.com
+IPFS_DOMAIN=gateway.originprotocol.com
+DISCOVERY_SERVER_URL=https://discovery.originprotocol.com
+DISCOVERY_SERVER_PORT=443
+IPFS_API_PORT=5002
+IPFS_GATEWAY_PORT=443
+IPFS_GATEWAY_PROTOCOL=https
--- a/dev.env
+++ b/dev.env
@@ -1,7 +1,6 @@
 BRIDGE_SERVER=https://bridge.originprotocol.com
 IPFS_DOMAIN=gateway.originprotocol.com
-DISCOVERY_SERVER_URL=https://discovery.originprotocol.com
-DISCOVERY_SERVER_PORT=443
+DISCOVERY_SERVER_URL=https://discovery.originprotocol.com:443
 IPFS_API_PORT=5002
 IPFS_GATEWAY_PORT=443
 IPFS_GATEWAY_PROTOCOL=https
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -2,17 +2,24 @@ const chalk = require('chalk')
 const startGanache = require('./helpers/start-ganache')
 const buildContracts = require('./helpers/build-contracts')
 const deployContracts = require('./helpers/deploy-contracts')
+const preBuild = require('./helpers/pre-build')
 const startIpfs = require('./helpers/start-ipfs')
 const startTestServer = require('./helpers/start-test-server')
 const watch = require('node-watch')
 const webpack = require('webpack')
-const webpackConfig = require('../webpack.config.js')
 
 const args = process.argv.slice(2)
 const shouldWatch = args.length && args[0] === 'serve'
 const noGanache = args.length && args[1] === 'no-ganache'
 
 const start = async () => {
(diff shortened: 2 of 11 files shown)


src/index.js
```diff
@@ -18,7 +18,7 @@ class Origin {
     ipfsGatewayPort = process.env.IPFS_GATEWAY_PORT,
     ipfsGatewayProtocol = process.env.IPFS_GATEWAY_PROTOCOL,
     attestationServerUrl = `${process.env.BRIDGE_SERVER}/api/attestations`,
-    discoveryServer = process.env.DISCOVERY_SERVER_PORT,
+    discoveryServer = process.env.DISCOVERY_SERVER_URL,
```
franckc: If we are using URL, do we still need discoveryServerPort ? Perhaps let's get rid of it  to make things simpler. So then for local development discoveryServer would point to something like "http://localhost:4000"

sparrowDom: Yup agree no need for 2 properties

sparrowDom:
@franckc I have added a pre-build step that copies contents of `dev.env` to `.env` in case the latter does not exist. So that `origin-box` users should have no problems and no additional manual work is required for the project to build. This is also important so that Continuous Integration (travis) can run as it needs the `.env` file. 

Without `origin-box` users will have to do this manually. 

Can you do another pass please?

dev.env
```diff
@@ -0,0 +1,6 @@
+BRIDGE_SERVER=https://bridge.originprotocol.com
+IPFS_DOMAIN=gateway.originprotocol.com
+DISCOVERY_SERVER_URL=https://discovery.originprotocol.com:443
```
franckc: No need to have :443 since this is the default for https

scripts/build.js
```diff
@@ -2,17 +2,24 @@ const chalk = require('chalk')
...
 
 const args = process.argv.slice(2)
 const shouldWatch = args.length && args[0] === 'serve'
 const noGanache = args.length && args[1] === 'no-ganache'
 
 const start = async () => {
+  console.log(chalk`\n{bold.hex('#26d198') ⬢  STARTING BUILD }\n`)
+  preBuild()
+
+  /* We need to load webpack config after preBuild step, because if .evn file does not exist pre-build
```
franckc: super nit: typo .evn vs .env

src/index.js
```diff
@@ -9,25 +9,16 @@ import Token from './resources/token'
...
-const defaultDiscoveryServerPort = '443'
-const defaultIpfsApiPort = '5002'
-const defaultIpfsGatewayPort = '443'
-const defaultIpfsGatewayProtocol = 'https'
-const defaultAttestationServerUrl = `${defaultBridgeServer}/api/attestations`
 const VERSION = require('.././package.json').version
 
 class Origin {
   constructor({
-    ipfsDomain = defaultIpfsDomain,
```
franckc: I believe this will break our production deployment when the DApp is served via IPFS because in that case no environment variable is defined. So we must keep those default assignments...

scripts/build.js
```diff
@@ -2,17 +2,24 @@ const chalk = require('chalk')
...
 const webpack = require('webpack')
-const webpackConfig = require('../webpack.config.js')
 
 const args = process.argv.slice(2)
 const shouldWatch = args.length && args[0] === 'serve'
 const noGanache = args.length && args[1] === 'no-ganache'
 
 const start = async () => {
+  console.log(chalk`\n{bold.hex('#26d198') ⬢  STARTING BUILD }\n`)
+  preBuild()
```
franckc: Will this break origin-box setup ?
See https://github.com/OriginProtocol/origin-box/blob/master/docker-compose.yml#L117

franckc:
@tomlinton I thin you have a much better understanding of origin-js/DApp configuration than I do at this point. Could you take a quick look and provide any comments ? Thanks !

sparrowDom:
This was a bad idea lets cancel it 😇

PR #433, opened 2018-09-10
📙 Discovery - Update listener/apollo to use new schema format

DanielVF:
Update listener/apollo to use new schema format.

- Name is now title
- Price is now a currency and an amount
- Index category
- Index subcategory

Should not affect the current dapp searching-by-description code


daemon/indexing/apollo/index.js
```diff
@@ -255,7 +255,12 @@ const resolvers = {
       return listing.category
     },
     price(listing) {
-      return {currency: 'ETH', amount: listing.priceEth}
+      if(listing.priceCurrency === undefined
```
franckc: The schema enforces that both of those are set and therefore origin-js marketplace.getListing should never return a Listing object that does not have those set. But agreed better to be paranoid and have those checks at different layer of the stack.

daemon/indexing/apollo/index.js
```diff
@@ -246,7 +246,7 @@ const resolvers = {
       return { walletAddress: 'S_WADDR' }
     },
     title(listing) {
-      return listing.name
+      return listing.title || "Untitled"
```
franckc: Would it make sense to actually return an empty string, so that the Dapp can differentiate a listing with no title vs with title ?

Now, note that the schema enforces a title is set since it is a "required" field: https://github.com/OriginProtocol/origin-js/blob/master/src/schemas/listing-core.json#L178
So this should never happen. But agreed better to be paranoid...

daemon/indexing/apollo/index.js
DanielVF: Given that origin.js now guarantees us a title, I'll remove this check.

sparrowDom: Also apollo schema guarantees us that

DanielVF: Well the Apollo schema crashes that part of the request if it's empty, which is what I was trying to avoid.

sparrowDom: Aha yes you are right... the `origin-js` guarantees the object validity at indexing time while apollo schema only at query time.

sparrowDom: Aha yes you are right... the `origin-js` guarantees the object validity at indexing time while apollo schema only at query time.

PR #438, opened 2018-09-11
Speedup for contractService.currentAccount

DanielVF:
If there’s already a defaultAccount set, there’s no reason to first make a remote call and then use the defaultAccount anyway.

### Checklist:


DanielVF:
Thanks for fixing that!

PR #443, opened 2018-09-11
📙 Discovery - Add generic GraphQL query method

DanielVF:
Add generic GraphQL query method.

This changes the response of discovery.search to a pure object with the results, rather than that fetch specific object - I'll make another PR to update the dapp to match it.


src/resources/discovery.js
```diff
@@ -5,6 +5,23 @@ class Discovery {
     this.fetch = fetch
   }
 
+  async query(graphQlQuery){
+    const url = `${this.discoveryServer}:${this.discoveryServerPort}`
```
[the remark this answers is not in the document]
DanielVF: Yeah, but the rest of the code isn't doing it yet, so I kept this PR focused on the just the query feature.

We can just make a new PR to switch it over, whenever we merge those url changes.

src/resources/discovery.js
```diff
@@ -5,6 +5,23 @@ class Discovery {
...
+        query: graphQlQuery
+      }),
+      headers: {
+        'Content-Type': 'application/json'
+      }
+    })
+    if(resp.status != 200){
+      throw Error('Got non-sucess code from GraphQL server')
+    }
+    return await resp.json()
```
sparrowDom: @DanielVF I think it would be better here for `origin-js` to return response object instead of JSON. A lot of things can go wrong for different reasons when issuing a fetch and DAPP might want to handle these differently (discovery server not found, response malformed etc.). 

When used by developers this way only a general "Got non-sucess code from GraphQL server'" error is thrown and a developer has no idea of knowing what went wrong. 

I am making a core filters implementation Pull Request and will change that there if you agree?

sparrowDom: Or alternatively we could throw different exceptions for different errors from `origin-js` too I guess.

DanielVF: Yes, I'd go for throwing different errors.

In general, I the user facing side of the consuming the this service can't really do much about failures. Either the search worked, or it didn't. And if it doesn't, we aren't really planning on giving the user detailed error messages, since there's nothing they can do.

I think it's also good not to couple the dapp to a specific GraphQL client implementation. We may want to swap it out later.

This also simples the dapp side code, since the logic to check for the details of errors no longer lives in each location discover is used. (We may still have code handling errors in DAPP, but that code no longer needs to duplicate how to find out that an error occurred.)

sparrowDom: Yeah I agree that throwing different exceptions is probably the best way to go since it hides the implementation details. 

On the other hand I think throwing different errors is important, since external and internal developers using `origin-js` should have quality feedback about what went wrong. Was JSON malformed, can discovery server not be reached, did discovery server return unexpected results... 

This way debugging `origin-js` to find the origin of the problem is not needed since the exception object is descriptive. 

And if need be we can communicate different errors to the user, but I agree that a general error that "something went wrong" is enough for the user.

PR #444, opened 2018-09-11
User registry upgradability

### LINKED ISSUE #369: User registry upgradability
Just like the marketplace, the user registry should be upgradable.

I propose that we follow the same structure that we use for the marketplace. We can have directories for each version of the user registry contract(s), and then adapters for each version. The resource will then tie all the adapters together.

tyleryasaka:

### Checklist:

### Description:

- Puts adapter structure in place for user registry upgradability
- Improves our adapter setup with the addition of resolvers (see explanation below)
- Adds more models

With resolvers, origin-js structure looks more like this:
- **Models**: High-level objects that contain *data* formatted for origin-js API consumers
- **Resources**: High-level objects that contain *methods* for the origin-js API; links ipfs and blockchain data (and maybe bridge server data in the future)
- **Resolvers**: High-level abstraction around our smart contracts with a clean API for interacting with the blockchain; links adapters together; consumed by resources
- **Adapters**: Objects with methods that implement contract-level details that change between contract versions; one adapter for each version

So basically what I did here was separate the concerns between the *resources* and the *resolvers*. Previously all of the heavy lifting was being done in the resources. This was problematic because:
1. It was really messy to be doing adapter resolution and data validation/linking to ipfs all in one place.
2. Multiple resources may sometimes wish to use the same contract abstractions provided by the adapters. Without separate resolvers, this would require duplicating code. (Example: attestations resource and users resource both need to access the user registry contracts.)

Resolves #369


DanielVF:
I really like it!

src/adapters/marketplace/_resolver.js
```diff
@@ -0,0 +1,284 @@
...
+  constructor({ contractService, ipfsService, store }) {
+    this.adapters = {
+      '000': new V00_MarkeplaceAdapter({ contractService, ipfsService, store }),
+      '001': new V01_MarkeplaceAdapter({ contractService, ipfsService, store })
+    }
+    this.versions = ['000', '001']
+    this.currentVersion = this.versions[this.versions.length - 1]
+    this.currentAdapter = this.adapters[this.currentVersion]
+    this.contractService = contractService
+    this.ipfsService = ipfsService
```
franckc: It does not seem ipfsService is used anywhere in this class. Can we get rid of it ?

src/adapters/marketplace/_resolver.js
```diff
@@ -0,0 +1,284 @@
...
+  Notification,
+  readStatus,
+  unreadStatus,
+  storeKeys
+} from '../../models/notification'
+
+class MarketplaceResolver {
+  constructor({ contractService, ipfsService, store }) {
+    this.adapters = {
+      '000': new V00_MarkeplaceAdapter({ contractService, ipfsService, store }),
```
franckc: Did we say that moving forward we would only use V01 marketplace ? And that V00 was only for testing purposes ?

src/resources/marketplace.js
```diff
@@ -289,36 +217,15 @@ class Marketplace extends Adaptable {
...
-        listingIndex,
-        offerIndex,
-        confirmationCallback
-      )
-    } else if (listingId) {
-      const { adapter, listingIndex } = this.parseListingId(listingId)
-
-      const ipfsHash = await this.listingIpfsStore.save(data)
-      const ipfsBytes = this.contractService.getBytes32FromIpfsHash(ipfsHash)
+    const ipfsHash = await this.reviewIpfsStore.save(data)
```
franckc: The type of data in case it is an offer vs a listing is different.
For an offer it is review data, but for listing it is TBD (not being used yet).
So the validation and upload code to IPFS will need to be conditional.
Or perhaps we should split this method into addListingData and addOffer data ?

src/adapters/marketplace/_resolver.js
tyleryasaka: Yeah. That just means that we will create all new listings with the V01 marketplace (which we are already doing.) If we don't include that adapter here in the resolver, it might as well not be in the codebase.

That said, I was thinking some more and it may just be simplest to not do our first release with unnecessary versions. Depending on how others feel, I may do a separate PR to remove the extra version.

cuongdo: +1 on removing V00. I don't see it adding anything, and it seems to be a source of ongoing confusion.

src/resources/marketplace.js
tyleryasaka: Oh, I see. Yeah I like the idea of splitting into 2 separate methods; will save that for a separate PR though. Used if/else to fix this for now.

src/adapters/marketplace/_resolver.js
tyleryasaka: Ok, will make a PR for that after this is merged

DanielVF: I actually think that having the V00 marketplace code is great. It forces us to aways be upgrade ready, and upgradable contract and js code is a part of everything we do and every release. Can't keep things more honest than that.

PR #446, opened 2018-09-12
make discovery server url a single parameter

sparrowDom:
### Description:

Dapp already uses a single parameter for discovery server url ( https://github.com/OriginProtocol/origin-dapp/blob/master/src/services/origin.js#L70 ) Lets just change it here also


franckc:
Nice ! :)

PR #447, opened 2018-09-12
Add pause & unpause to token_cli

cuongdo:
### Checklist:

- [ ] ~~Code contains relevant tests for the problem you are solving~~

### Description:

This adds the `pause` and `unpause` actions to the token CLI, which
pause and unpause token transfers. These may only be called by the token
contract owner.

This also adds `sendTransaction` to the token library, which sends an
Ethereum transaction and polls for the receipt. This is necessary
workaround for an issue where the normal `contract.methods.MyMethod.send()`
function never returns a transaction receipt for certain calls:

https://github.com/INFURA/infura/issues/95

Manually tested the token faucet on Rinkeby & localhost to ensure that it still
works.

### DIFF
--- a/token/lib/token.js
+++ b/token/lib/token.js
@@ -196,10 +196,14 @@ class Token {
     // `await contract.methods.MyMethod(...).send(...)`:
     //
     // https://github.com/INFURA/infura/issues/95
-    const maxSleep = 15 * 60 * 1000 // 15 minutes, in milliseconds
+
+    // Blocks are mined every ~15 seconds, but it sometimes takes ~40-60 seconds
+    // to get a transaction receipt from rinkeby.infura.io.
+    const maxSleep = 120000
     let totalSleep = 0
     let sleepTime = 1000
     while (totalSleep <= maxSleep) {
+      this.vlog(`waiting ${sleepTime / 1000}s for transaction receipt`)
       await sleep(sleepTime)
 
       if (transactionHash) {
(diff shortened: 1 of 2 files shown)


token/lib/token.js
```diff
@@ -100,6 +110,148 @@ class Token {
...
+      })
+
+    // Poll for the transaction receipt, with an exponential backoff. This works
+    // around some strange interactions between web3.js and some web3 providers.
+    // For example, this issue sometimes prevents transaction receipts from
+    // being returned when simply calling
+    // `await contract.methods.MyMethod(...).send(...)`:
+    //
+    // https://github.com/INFURA/infura/issues/95
+    const maxSleep = 15 * 60 * 1000 // 15 minutes, in milliseconds
```
franckc: Wow. 15min is a long time ? I think on average blocks are mined every 15sec.
But I guess it does not really hurt to wait.

token/lib/token.js
```diff
@@ -100,6 +110,148 @@ class Token {
...
+            this.vlog('transaction successful')
+            return receipt
+          } else {
+            throw new Error('transaction failed')
+          }
+        } else {
+          this.vlog('waiting for transaction receipt')
+        }
+      } else {
+        this.vlog('waiting for transaction hash')
```
franckc: Nit: maybe log how many sec the next check will be performed. So that the operator running the script does not think it is hanging.

token/lib/token.js
cuongdo: Yes, good point. Fixed.

token/lib/token.js
cuongdo: Good point. I've removed this log message and added another at the beginning of the `while`.

PR #448, opened 2018-09-12
🚜 Sample data generator

DanielVF:
Create user profiles, sample listings, and sample offers.

Make as many as you want! Load test your app!

Currently made to run against a local truffle environment, but should not be hard to change to work on a testnet.


DanielVF:
(Fixing the eslint issue)

DanielVF:
@tyleryasaka I really don't want to add large dependency just for this. Faker is 14mb zipped, and I'd rather keep our list of dependencies low for security reasons as well. One less way developers can get attacked.

In this case, it's just a list of words so it's not going to need to be upgraded. The words used to not take up that much space in the file -- until prettier wanted to put each word on it's own line. :) The only disadvantage is that you have to scroll before editing the file.

tyleryasaka:
@DanielVF Yeah Faker is big, but we would just add it as a dev dependency. That won't affect the size of our production build and won't introduce any security risks in production. Not a big deal though, I'm fine with this list.

As far as the words being inline - how about we move the list of words to a separate json file and import it?

DanielVF:
A separate file would be great - where you recommend we put it?

Dev dependencies are still possible security problems, since it's an attack vector on an blockchain developer, which is a nice route into other valuable things.

tyleryasaka:
Meh, `scripts/sample-data.json` would work for me. Or wherever you think it would make sense.

sparrowDom:
Using this to test search functionality. Works perfectly 💪

scripts/sample-data.js
```diff
@@ -0,0 +1,145 @@
...
+// 3) Users make offers on other's listings
+
+// Local Usage:
+// scripts/sampleData.js <number of listings> <number of offers>
+// docker exec -ti origin-js node scripts/sampleData.js 42 17
+
+const NUMBER_OF_LISTINGS = parseInt(process.argv[2] || '30')
+const NUMBER_OF_OFFERS = parseInt(process.argv[3] || NUMBER_OF_LISTINGS / 2)
+
+const NAMES = [
```
franckc: I want in !!! :)

scripts/sample-data.js
```diff
@@ -0,0 +1,145 @@
...
+  return word.charAt(0).toUpperCase() + word.slice(1)
+}
+
+/*
+ * Setup
+ */
+
+const web3Provider = new Web3.providers.HttpProvider('http://localhost:8545')
+const web3 = new Web3(web3Provider)
+const ipfsUrl = urllib.parse('http://origin-js:8080')
```
franckc: super nit: maybe add a comment explaining that host "origin-js" is the virtual host created by the origin-js container when it runs in origin-box.

scripts/sample-data.js
```diff
@@ -0,0 +1,145 @@
...
+    )
+    const newListing = await o.marketplace.createListing({
+      schemaVersion: '1.0.0',
+      listingType: 'unit',
+      title: listingName,
+      category: CATEGORIES[Math.floor(Math.random() * CATEGORIES.length)],
+      subCategory: CATEGORIES[Math.floor(Math.random() * CATEGORIES.length)],
+      language: 'en-US',
+      description: description,
+      price: { currency: 'ETH', amount: price.toString() },
```
franckc: Feature request (can totally be done as separate PR): let's add a commission.
Something like:
`commission: { currency: 'OGN', amount: <same random amount> }`
That will allow us to test that ranking boosts listings with higher commission.

franckc:
ah. the build fails because:
```
/home/travis/build/OriginProtocol/origin-js/scripts/sample-data.js
  36:44  error  Strings must use singlequote  quotes
```

scripts/sample-data.js
```diff
@@ -0,0 +1,145 @@
...
+  'Daniel',
+  'Matthew',
+  'Kay',
+  'Пётр',
+  '계룡/鷄龍',
+  'Urist'
+]
+const CATEGORIES = ['Housing', 'Collectibles', 'For Sale', 'Services']
+const wordsFilePath = path.join(__dirname, "sample-words.json")
+const WORDLIST = JSON.parse(fs.readFileSync(wordsFilePath))
```
tyleryasaka: You can actually just import the json file.

```js
import WORDLIST from './sample-words.json'
```

DanielVF: I don't think node natively supports imports without special flags. I've switched it to require the words list file though.

sparrowDom:
@DanielVF I hope you don't mind... I have added the commission into the PR since I need it to test search... Also fixed a small bug in listeners. Looks like `adapters` have moved from `marketplace` to `marketplace.resolvers`

DanielVF:
Great. Do all the default accounts have starting OGN now? Otherwise does it break when you include a commission amount, but have no OGN?

sparrowDom:
Looks like they do, because the listings get added to blockchain and the indexing listener picks them up and indexes them

franckc:
Yes the migration scripts give 200 OGN to all accounts. See this [script](https://github.com/OriginProtocol/origin-js/blob/master/contracts/migrations/6_transfer_tokens_for_dev.js).

franckc:
@DanielVF @sparrowDom 
Going to merge this in a few minutes unless there is any objection...

scripts/sample-data.js
tyleryasaka: We're already doing it. https://github.com/OriginProtocol/origin-js/blob/master/src/resources/users.js#L3

scripts/sample-data.js
```diff
@@ -0,0 +1,144 @@
...
+    console.log(
+      chalk` ⬢  Creating listing {bold.hex('#d408f4') ${listingName}} from {bold.hex('#09f4a6') ${userFirst}}`
+    )
+
+    const newListing = await o.marketplace.createListing({
+      schemaVersion: '1.0.0',
+      listingType: 'unit',
+      title: listingName,
+      category: CATEGORIES[Math.floor(Math.random() * CATEGORIES.length)],
+      subCategory: CATEGORIES[Math.floor(Math.random() * CATEGORIES.length)],
```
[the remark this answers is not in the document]
sparrowDom: @DanielVF bump (in case you overlooked the comment)

PR #450, opened 2018-09-12
Support boost approval and commissions

jordajm:
### Checklist:

- This will need to have some unit tests added to it. Also, `ClaimHolderPresigned` test is failing - not sure why.

### Description:

Please explain the changes in this PR:

Supports https://github.com/OriginProtocol/origin-dapp/pull/491

Adds two new methods to the token API:
- `approveContract()`
- `getAllowance()`


src/resources/token.js
```diff
@@ -54,6 +58,62 @@ class Token extends ResourceBase {
     await this.getContract()
     return await this.contract.methods.paused().call()
   }
+
+  /**
+   * Relays calls to the contract service
+   */
+  async call(methodName, args, opts) {
```
micahalcorn: Can this not use [`contractService.call`](https://github.com/OriginProtocol/origin-js/blob/john/boost_approval/src/services/contract-service.js#L187)?

micahalcorn: Why does this not call the `contractService` method directly?

micahalcorn: Ah, I see, it needs the `contractName`. ✅

jordajm: Right on - it's modeled after the marketplace adapter here:
https://github.com/OriginProtocol/origin-js/blob/john/boost_approval/src/adapters/marketplace/v00.js#L26-L33

src/adapters/marketplace/v00.js
```diff
@@ -39,10 +39,15 @@ class V00_MarkeplaceAdapter {
...
     ipfsBytes,
-    { deposit = '0', arbitrator },
+    { deposit = '0', arbitrator, commission },
     confirmationCallback
   ) {
     const from = await this.contractService.currentAccount()
+    deposit = commission && commission.amount || '0'
+
+    if (deposit !== '0') {
+      deposit = (Number(deposit)*10**18).toString()
```
franckc: In general I think it's not safe to use JS Number for such large values... Let's use BigNumber everywhere ?

Either by using the [Big Number](https://www.npmjs.com/package/big-number) package directly (we'll need to add it to package.json) or via https://web3js.readthedocs.io/en/1.0/web3-utils.html#tobn - I'd lean towards the former.

src/resources/token.js
```diff
@@ -16,6 +16,10 @@ class Token extends ResourceBase {
     this.getTokenAddress = async function() {
       return await marketplace.getTokenAddress()
     }
+
+    this.contractService = contractService
+    this.marketplace = marketplace
+    this.contractName = 'OriginToken'
```
franckc: As opposed to hardcoding it, I think it is accessible via OriginTokenContract.contractName

src/resources/token.js
```diff
@@ -54,6 +58,62 @@ class Token extends ResourceBase {
...
+  async getAllowance(tokenOwnerAddress) {
+    await this.getContract()
+    await this.setMarketplaceContractAddress()
+
+    const allowanceRemaining = await this.call(
+      'allowance',
+      [tokenOwnerAddress, this.marketplaceContractAddress],
+    )
+    
+    const allowanceNum = parseFloat(allowanceRemaining)
```
franckc: Same comment as above. Let's use BigNumber.

src/adapters/marketplace/v00.js
```diff
@@ -39,10 +39,15 @@ class V00_MarkeplaceAdapter {
 
   async createListing(
     ipfsBytes,
-    { deposit = '0', arbitrator },
+    { deposit = '0', arbitrator, commission },
     confirmationCallback
   ) {
     const from = await this.contractService.currentAccount()
+    deposit = commission && commission.amount || '0'
```
franckc: Paranoia: should we check commission.currency==='OGN' and add a comment that in V1 we only support commission in that currency ?

micahalcorn:
Closing this in favor of #475 😅

PR #451, opened 2018-09-13
Generalize IpfsDataStore and implement OfferAccept and ListingWithdraw data types

franckc:
### Checklist:

### Description:
- Generalized further the IPFS data storage abstraction by implementing an "IpfsDataStore" service that can be used to store/load/validate any data type.
- Replaced the schemaVersion field in the data with schemaId which contains both the data type and the version. This seems more general and allows to do extra validation on the data being manipulated.
- Used the new IpfsDataStore to implement 2 additional data types: OfferAccept and ListingWithdraw. It is a little odd but currently those objects do not store any data - these are placeholder for future use.


test/resource_marketplace.test.js
```diff
@@ -184,7 +184,7 @@ describe('Marketplace Resource', function() {
       expect(notifications[0].type).to.equal('seller_listing_purchased')
       expect(notifications[0].status).to.equal('unread')
 
-      await marketplace.acceptOffer('999-001-0-0')
+      await marketplace.acceptOffer('999-001-0-0', {})
```
tyleryasaka: I think we should use default parameters so that we don't have to pass in empty objects. See other comment.

src/resources/marketplace.js
```diff
@@ -173,10 +174,8 @@ class Marketplace {
    * @return {Promise<{timestamp, transactionReceipt}>}
    */
   async acceptOffer(id, data, confirmationCallback) {
```
tyleryasaka: I think all of the `data` params should have default values so that they can be omitted.

E.g. `acceptOffer(id, data = {}, confirmationCallback)`

test/resource_marketplace.test.js
franckc: Sounds good. Done.

PR #458, opened 2018-09-14
Core filters functionality

sparrowDom:

### Checklist:

### Description:
- implement search filters that are currently required by the Dapp (category, subCategory, price)
- boost matching of the title field 
- create basic ranking when it comes to amount of `OGN` tokens in commission field 
- when returning a set of search results also return the max price in Ether of all the results. (required by the price filter)

###Dependant PRs: 
- https://github.com/OriginProtocol/origin-dapp/pull/503
- https://github.com/OriginProtocol/origin-box/pull/52


daemon/indexing/apollo/index.js
```diff
@@ -97,6 +97,7 @@ const typeDefs = gql`
     # itemsPerPage: Int!
     # totalNumberOfPages: Int!
     nodes: [Listing]
+    maxPrice: Float
```
franckc: Couple questions:
 - In the future there could be other aggregated fields we want to query. For ex min price, num bedrooms, etc... So should this be wrapped in an object. Like ListingStats ?
 - Since this aggregated data is used to display the range on a filter, don't we need to get that information *before* making the search query, at the time we display the filter ? If so, should this be a separate query with its own graphql entrypoint. Like listingStats and it would return a ListingStats object ?

daemon/indexing/apollo/index.js
```diff
@@ -195,7 +196,7 @@ const typeDefs = gql`
 
   # The "Query" type is the root of all GraphQL queries.
   type Query {
-    listings(searchQuery: String, filters: [ListingFilter!]): ListingPage, #(searchQuery: String, page: Page, order: ListingOrder, filters: [ListingFilter!]): ListingPage,
+    listings(searchQuery: String, category: String, filters: [ListingFilter!]): ListingPage,
```
franckc: Why is category a string vs added as a ListingFilter ?
Also, we'll need to add filter by subCategory.

daemon/indexing/apollo/index.js
sparrowDom: - yeah good idea. Thinking of naming it stats so it is more general and could also apply to different index types once we implement it (people for example)?
- The search page that displays results always has a search query preceding it. So filters displayed on that page can populate their properties from search response. The initial query a user makes is from the navigation bar (general search) and at that point the page displays no filters. So there is no need to pre-issue a search to discovery server.

daemon/indexing/apollo/index.js
sparrowDom: Yeah good point. I have changed it to be added as a `ListingFilter` like suggested. Subcategory is already added as a `ListingFilter`.

src/resources/discovery.js
```diff
@@ -27,11 +33,11 @@ class Discovery {
    * unexpected. To get JSON result caller should call `await searchResponse.json()` to get the
    * actual JSON.
    * @param searchQuery {string} general search query
-   * @param listingType {string} one of the supported listingTypes
+   * @param category {string} one of the supported categories
```
franckc: Remove. This param is deprecated and now part of filters.

daemon/indexing/apollo/index.js
franckc: Still there ?

PR #461, opened 2018-09-15
ERC725 Fixes

### LINKED ISSUE #170: ERC 735: Rename claimType to topic
[Fabian has requested](https://github.com/ethereum/EIPs/issues/735) that implementations of ERC735 change the name of `claimType` to `topic`, which is a more precise (and less confusing) name. We will want to change the name used in our [smart contracts](https://github.com/OriginProtocol/origin-js/blob/master/contracts/contracts/identity/ClaimHolder.sol#L13) as well as any variables that reference this value across our repos (including the JS in this repo and the Python in the bridge server).

tyleryasaka:
### Checklist:

### Description:

Various ERC725 fixes. See commit messages for details 😃 

Resolves #170

### DIFF
--- a/contracts/contracts/identity/KeyHolderLibrary.sol
+++ b/contracts/contracts/identity/KeyHolderLibrary.sol
@@ -147,6 +147,10 @@ library KeyHolderLibrary {
       public
       returns (bool success)
   {
+      if (msg.sender != address(this)) {
+        require(keyHasPurpose(_keyHolderData, keccak256(abi.encodePacked(msg.sender)), 1), "Sender does not have management key"); // Sender has MANAGEMENT_KEY
+      }
+
       require(_keyHolderData.keys[_key].key == _key, "No such key");
       emit KeyRemoved(_key, _purpose, _keyHolderData.keys[_key].keyType);
 
--- a/contracts/contracts/identity/KeyHolderLibrary.sol
+++ b/contracts/contracts/identity/KeyHolderLibrary.sol
@@ -9,7 +9,7 @@ library KeyHolderLibrary {
   event Approved(uint256 indexed executionId, bool approved);
 
   struct Key {
-      uint256 purpose; //e.g., MANAGEMENT_KEY = 1, ACTION_KEY = 2, etc.
+      uint256[] purposes; //e.g., MANAGEMENT_KEY = 1, ACTION_KEY = 2, etc.
       uint256 keyType; // e.g. 1 = ECDSA, 2 = RSA, etc.
       bytes32 key;
   }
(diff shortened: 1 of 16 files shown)


contracts/contracts/identity/KeyHolderLibrary.sol
```diff
@@ -143,14 +143,43 @@ library KeyHolderLibrary {
...
+        require(keyHasPurpose(_keyHolderData, keccak256(abi.encodePacked(msg.sender)), 1), "Sender does not have management key"); // Sender has MANAGEMENT_KEY
+      }
+
       require(_keyHolderData.keys[_key].key == _key, "No such key");
-      emit KeyRemoved(_keyHolderData.keys[_key].key, _keyHolderData.keys[_key].purpose, _keyHolderData.keys[_key].keyType);
+      emit KeyRemoved(_key, _purpose, _keyHolderData.keys[_key].keyType);
+
+      // Remove purpose from key
+      uint256[] storage purposes = _keyHolderData.keys[_key].purposes;
+      for (uint i = 0; i < purposes.length; i++) {
```
nick: I originally had this as `purpose` instead of `purposes` in order to avoid these kind of loops and extra required storage on-chain (and associated gas costs). My idea was that you can represent multiple purposes with a single uint256 with each bit representing a different purpose, allowing us to represent 128 purposes for a single key. Since we only have 4 possible purposes right now, I figured 128 would be plenty. I also used this technique in the Marketplace contract to represent [multiple resolutions](https://github.com/OriginProtocol/origin-js/blob/master/contracts/contracts/marketplace/v01/Marketplace.sol#L270). Would be interested in your thoughts on that approach. I know it's different to the current standard so I should probably bring it up there too.

tyleryasaka: Yeah I don't see why that can't be done to improve efficiency. I'm still thinking through the standard and trying to form my own opinions.

I'm not opposed to proposing a change to the standard. However, I do think we should be compliant with whatever the current standard is.

tyleryasaka:
Ok, sounds good. I'll go ahead and merge. These changes don't reflect my opinion on how things should be; they're just to get our contracts compliant with the current standard. I would definitely like to have further discussions around proposing possible changes to the standard.

contracts/contracts/identity/KeyHolderLibrary.sol
joshfraser: I really liked @nick's proposal of using bitmasks for this. We should try and get it into the standard. Do we really think we need 128 different purposes? Seems like an uint8 ought to be enough.

joshfraser: Ah, didn't know that, but it makes sense. Two uint128's seems more than adequate to me.

PR #472, opened 2018-09-17
🚜 Sample data subcategories

DanielVF:
- Sample data generator now uses actual subcategories
- Tiny refactoring by adding randomPick function


DanielVF:
I'm going to change the categories to the untranslated ones.

franckc:
ah yes, thanks for doing that @DanielVF - then I'll merge into master...

scripts/sample-data.js
```diff
@@ -30,10 +30,18 @@ const NAMES = [
   '계룡/鷄龍',
   'Urist'
 ]
-const CATEGORIES = ['Housing', 'Collectibles', 'For Sale', 'Services']
+const CATEGORIES = {
+  forSale: ['artsCrafts', 'farmGarden', 'heavyEquipment', 'tickets'],
```
sparrowDom: Double checked now and I see translated listings stored in elasticsearch index:
screenshot 2018-09-18 19 02 28

screenshot 2018-09-18 19 04 25

Has anybody already thought through how we should handle localisation? @DanielVF @franckc

franckc: @sparrowDom Could these be old documents that were index prior to Daniel's change ?
Regarding localization I don't think we should localize cat/subcat when indexing. These are just to be used as filters for searches.

sparrowDom: Yeah @franckc you are right. It makes no sense to localize them, since it is only going to work for english language. I guess we can solve that once we sort out localisation within our schemas. 

@DanielVF maybe just change them to full translation strings like `schema.forSale.antiques` instead of `antiques` https://github.com/OriginProtocol/origin-dapp/blob/master/public/schemas/for-sale.json#L27-L70

scripts/sample-data.js
```diff
@@ -30,10 +30,18 @@ const NAMES = [
   '계룡/鷄龍',
```
sparrowDom: While this PR is still open can you also change the line 16 (can not comment up there :) ) to: `docker exec -ti origin-js node scripts/sample-data.js 42 17`

DanielVF:
This is because the accounts seem to only have 200 OGN now?

Thus randomly trying to spend 0-1000 OGN fails most of the the time. Locally I just just have it set to spend 0-20 OGN, while I waited to see if that was a perma-change.

sparrowDom:
Aha I see, any idea where can I set it to less?

micahalcorn:
I was just talking to @cuongdo about changing the test accounts to have varying amounts of token instead of [200](https://github.com/OriginProtocol/origin-js/blob/master/contracts/migrations/6_transfer_tokens_for_dev.js). I often want to test with 0️⃣ OGN in the wallet, and sometimes with other amounts.

micahalcorn:
I have the following use cases:

- account with none
- account with less than the boost I want (currently under 500, but may need to be lower)
- account with a realistic amount (not 999,996,400)

Somewhat unrelated, but I also miss the days when we had sample listings belonging to more than just account 2 and also sample purchases in various states.

DanielVF:
@sparrowDom line 98:

```
const commission = Math.floor(Math.random() * 10)
```

DanielVF:
@micahalcorn Yeah, I agree that it would be very useful to prepopulate the dev environment with a standard set of listings in every state, and involving more than just two buyers.

scripts/sample-data.js
jordajm: @sparrowDom @franckc - I think in our meeting about schemas a few weeks ago, we decided not to try to localize any strings in `origin-js`, and that we would only localize strings in the schemas in `origin-dapp`. I'm not sure if the localization setup that we currently have would support translation of strings in `origin-js` but I can look into it if we're now considering it.

sparrowDom: @jordajm  What I meant by localisation in schemas is not that we would translate part of schemas using localisation module, but that our schemas support different languages. (sorry for lack of clarity)

Currently there is no way of storing the same listing (with the same id) in different languages. We have discussed at some point that language specific fields (like title, description ....) will be in an array of objects where each object would have a corresponding language. 

And once we support that, we will also need to change indexing/search engine quite a lot, and we can figure out then how to index data. 

But for now, that we just keep untranslated keys (like `schema.forSale.antiques`) that define schema `cateogry` and `subCategory`

franckc:
@sparrowDom Are you ok with the state of this PR or do you still think changes are required ? If you are ok then could you change your review status to accepted so that I can merge it ? Thanks :)

sparrowDom:
@franckc yup its ok (have pushed the two minor changes in)

DanielVF:
@sparrowDom I'm not seeing the Dapp use those whole translation strings for the subcategories? Are you seeing something different?

I just created another listing in the dapp, and I see `..."category":"forSale","subCategory":"carsTrucks"...`

sparrowDom:
@DanielVF yeah I see we have a nice mess here:
- listings created by the Dapp have format (like you specify) `..."category":"forSale","subCategory":"carsTrucks"...` (non translated, not full translation key)
- listings (~6) that are present when you install the Dapp have format 
`... "category" : "For Sale", "subCategory" : "Cars & Trucks"...` (translated)

and I think none of them is correct, since the translation module requires the whole string to translate category/subcategory. See example: https://github.com/OriginProtocol/origin-dapp/blob/master/public/schemas/for-sale.json#L27-L70

Since it will be hard to change listings in IPFS once we go live we should correct this (and write the whole translation key to IPFS). Except if I am getting something wrong here? @franckc @jordajm

DanielVF:
I can confirm that your way matches how the dapp displays categories. The dapp itself is not consistent as you said. I believe we should go with the full string in the subcategory. We can **merge this PR** and I'll:

- [ ] update the sample listings
- [ ] update the way the dapp saves subcategories

sparrowDom:
👍 sounds great to me

franckc:
@DanielVF Is this ready to get merged into master ? Pls let me know and I'll do the honor...

DanielVF:
Should be ready to merge. I threw in some other fixes that had made it stop working in our setup.

PR #477, opened 2018-09-18
Offer Commission Repair

micahalcorn:
### Checklist:

### Description:

Without this, offers cannot be made for boosted listings.

### DIFF
--- a/src/adapters/marketplace/v00.js
+++ b/src/adapters/marketplace/v00.js
@@ -100,27 +100,40 @@ class V00_MarkeplaceAdapter {
   }
 
   async makeOffer(listingId, ipfsBytes, data, confirmationCallback) {
-    const {
-      finalizes,
-      affiliate,
-      commission,
-      price,
-      arbitrator,
-      currency
-    } = data
+    const { affiliate, arbitrator, commission = {}, finalizes, totalPrice = {}, unitsPurchased } = data
+    // For V1, we only support quantity of 1.
+    if (unitsPurchased != 1)
+      throw new Error(
+        `Attempted to purchase ${unitsPurchased} - only 1 allowed.`
+      )
+
+    // Convert price to correct units for blockchain
+    let currency, price
+    if (totalPrice.currency === 'ETH') {
+      currency = 'ETH'
+      price = this.contractService.web3.utils.toWei(totalPrice.amount, 'ether')
+    } else {
+      // handle ERC20
+      currency = this.contractService.currencies[totalPrice.currency]
+      // TODO consider using ERCStandardDetailed.decimals() (for tokens that support this) so that we don't have to track decimals ourselves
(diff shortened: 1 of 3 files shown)


src/resources/marketplace.js
```diff
@@ -217,6 +217,7 @@ class Marketplace {
       ipfsBytes,
       {
         price,
+        commission: offerData.commission,
```
franckc: Everything passed in offerData gets stored in IPFS. so if we pass the commission this way, let's use a Money object with {currency, amount). and then call the contract with
`commission: offerData.commission.amount`

An alternative would be that the  DApp does not pass the commission. It is up to origin-js to set it before calling the contract. This method would load the listing ipfs data and get commission from there. When we implement multi-unit and fractional usage, I think we might have to do that anyway to calculate the commission (for example to split it in the case of multi-units so that it gets spread over N units ?).

Super nit: move commission one line down to keep price and currency together since they are related.

micahalcorn:
@franckc I made some changes in 62e67c70df6051565653d226d24b036af549d685 to handle a commission object (now passed in https://github.com/OriginProtocol/origin-dapp/pull/514) and also took the opportunity to move the logic to the adapter in an effort to conform to the convention set by the other methods. There may be a reason why this stuff was all in the resource before.

Also, I realized that I had completely forgotten that @tyleryasaka was going to work on this, as he and I had discussed just moments before I jumped on this issue yesterday. 😑 

So, these changes work in the DApp, but the tests are now failing. Tyler, if you want to just scrap this PR and run with your stashed code, feel free to do that. Otherwise, I'll continue to investigate what's wrong here.

src/adapters/marketplace/v00.js
```diff
@@ -100,27 +100,40 @@ class V00_MarkeplaceAdapter {
...
+    if (totalPrice.currency === 'ETH') {
+      currency = 'ETH'
+      price = this.contractService.web3.utils.toWei(totalPrice.amount, 'ether')
+    } else {
+      // handle ERC20
+      currency = this.contractService.currencies[totalPrice.currency]
+      // TODO consider using ERCStandardDetailed.decimals() (for tokens that support this) so that we don't have to track decimals ourselves
+      // https://github.com/OpenZeppelin/openzeppelin-solidity/blob/6c4c8989b399510a66d8b98ad75a0979482436d2/contracts/token/ERC20/ERC20Detailed.sol
+      const currencyDecimals = currency && currency.decimals
+      price = String(Number(totalPrice.amount) * 10**currencyDecimals)
```
franckc: Doesn't this have the risk to overflow if we use regular JS Number vs BigNumber ?
Won't cause issue in V1 since this code path won't get exercised as we only support listing prices in ETH. But might break in future ?

micahalcorn: Good call @franckc 👍 I just copied that over but will use the `toWei` like I did in the listing logic.

tyleryasaka: @franckc ah ok, I think you're right. I'll fix this. Thanks.

PR #480, opened 2018-09-18
Make configurable via environment

tomlinton:

### Checklist:

### Description:

- Make postgres and elasticsearch configurable via environment variables rather than use hardcoded values
- Use published origin-js rather than relative build file

These are decoupled from `origin-js` if this gets merged. I know I said we should leave these here for now @franckc but I think we should probably split them out into two separate packages at the top of `origin-box`. They are sharing the same `package.json` but not all the dependencies are common. I propose:

- GraphQL server moves to `origin-box/origin-discovery`
- Event listener moves to `origin-box/event-listener`

They each get their own `package.json`. Migrations can live with either one. It should be quick and easy to do this unless @DanielVF or @sparrowDom are working on this code in branches?


daemon/indexing/listener/listener.js
```diff
@@ -1,9 +1,9 @@
 const fs = require('fs')
 const http = require('http')
 const urllib = require('url')
+const Origin = require('origin')
```
DanielVF: While this is right in the long term, this change would break things right now, since the NPM origin.js isn't up to date. Will NPM link work here? Where does it need to happen? In the container?

tomlinton: Yea that is right. I'll use `npm link` to link it to a version of `origin-js` that I build. You can see an [example here](https://github.com/OriginProtocol/origin-box/blob/master/docker-compose.yml#L167) for `origin-dapp`. It lets me put the container setups in place without worrying about maintaining that relative import and I don't have to build `origin-js` in each of the containers for the event listener and apollo server.

sparrowDom: Should it be in a separate container like dapp is, and wait for `origin-js` to build the same way dapp does it?

tomlinton: Yep exactly, that is what I'm going to do. You can see my initial pass [here](https://github.com/OriginProtocol/origin-box/commit/28514da3476ff2839dfcabe3d082864a645003d0).

tomlinton:
@franckc gentle reminder that I have to get this merged. :grin: 

I can't configure the event listener or apollo in any setup other than `docker-compose` without it.

franckc:
@tomlinton sorry ! rebasing your branch and merging this in asap...

PR #483, opened 2018-09-18
Money + Commission Fixes

tyleryasaka:
### Checklist:

### Description:

- Create moneyToUnits method on contract service which reads a money object and returns the blockchain units for it based on currency
- Use BigNumber for safely converting ERC20 currencies to blockchain units
- Fix bug where offer price validation was incorrect for ERC20 tokens
- Makes `Money` a model rather than util (seems more like a model to me)


micahalcorn:
I'll defer to @franckc on this but I should point out that [I relied on `toWei` for the deposit calculation](https://github.com/OriginProtocol/origin-js/commit/5ae22f636477b16af8103487cd7aca36c318daa6#diff-514425d04204d4428ab9f087ce363251R56). It may be more appropriate to use `token.decimals()` and actually do the math?

tyleryasaka:
@micahalcorn `toWei` works only when the number of decimals supported by the token is 18 (Eth has 18 decimals, one unit of which is a Wei).

I am using the same method that @franckc used here: https://github.com/OriginProtocol/origin-js/blame/master/token/lib/token.js#L22

This method works for other decimal amounts. We should only use `toWei` for eth.

tyleryasaka:
Went ahead and added my commission tests and fixes to this PR.

src/services/contract-service.js
```diff
@@ -229,6 +230,25 @@ class ContractService {
...
+      return Web3.utils.toWei(money.amount, 'ether')
+    } else {
+      const currency = this.currencies[money.currency]
+      // handle ERC20
+      // TODO consider using ERCStandardDetailed.decimals() (for tokens that support this) so that we don't have to track decimals ourselves
+      // https://github.com/OpenZeppelin/openzeppelin-solidity/blob/6c4c8989b399510a66d8b98ad75a0979482436d2/contracts/token/ERC20/ERC20Detailed.sol
+      const currencyDecimals = currency && currency.decimals
+      if (currencyDecimals) {
+        const scaling = BigNumber(10).exponentiatedBy(currencyDecimals)
+        return String(BigNumber(money.amount).multipliedBy(scaling))
```
franckc: how about using the BigNumber toString() method ?

src/adapters/marketplace/v00.js
```diff
@@ -100,41 +100,30 @@ class V00_MarkeplaceAdapter {
   }
 
   async makeOffer(listingId, ipfsBytes, data, confirmationCallback) {
-    const { affiliate, arbitrator, commission = {}, finalizes, totalPrice = {}, unitsPurchased } = data
+    const { affiliate, arbitrator, commission, finalizes, totalPrice = {}, unitsPurchased } = data
```
franckc: affiliate, arbitrator, finalizes will get saved in IPFS. We had discussed storing the blockchain data under a separate "_untrustworthy" object. So that developers don't get confused as to what is the source of truth between onchain and offchain data. Add a TODO here so that we remember ?

src/services/contract-service.js
tyleryasaka: doh 🤦‍♂️

src/adapters/marketplace/v00.js
tyleryasaka: 👍

test/resource_marketplace.test.js
```diff
@@ -182,6 +189,21 @@ describe('Marketplace Resource', function() {
...
+      let errorThrown = false
+      let errorMessage
+      try {
+        await marketplace.getOffer('999-001-1-0')
+      } catch(e) {
+        errorThrown = true
+        errorMessage = String(e)
+      }
+      expect(errorThrown).to.be.true
+      expect(errorMessage).to.equal('Error: Invalid offer: insufficient commission amount for listing')
```
franckc: FYI, you can also use the chaiAsPromised package for these types of tests that an async function throws.
See https://github.com/OriginProtocol/origin-js/blob/master/test/service-data-store.test.js#L163

tyleryasaka: Thanks. Yeah, I did look at that once. Can't remember if I had a good reason not to use it. Maybe a future refactor can use that.

PR #486, opened 2018-09-18
Fix weird webpack loop

cuongdo:
### Checklist:

### Description:

Caused by importing of _helper.js. Don't know why.

### DIFF
--- a/contracts/test-alt/TokenLib.js
+++ b/contracts/test-alt/TokenLib.js
@@ -1,6 +1,6 @@
 import assert from 'assert'
-import helper, { contractPath } from '../contracts/test-alt/_helper'
-import Token from '../token/lib/token'
+import helper, { contractPath } from './_helper'
+import Token from '../../token/lib/token'
 
 // These tests are for the token library that the token CLI uses. We don't
 // validate the effects of various operations. That is left to the contract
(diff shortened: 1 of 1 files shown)


tomlinton:
I think this is caused by the [hot module reload](https://github.com/OriginProtocol/origin-js/blob/3172a45035a9489627a6e9dcc3ef5f2c396a9414/scripts/helpers/start-test-server.js#L7) used to start webpack for the test server. I removed it and everything works OK except the test server no longer displays any tests when I open it. :confused: 

See https://github.com/webpack/webpack/issues/2157

franckc:
Interesting... Does anyone actually use the test server ? I find it really hard to use because when a test fails, it does not provide enough details as to what the failure is. I'd personally be inclined to get rid of it...

tomlinton:
It uses a bunch of resources in `origin-box` because everything gets compiled twice so I'm all for getting rid of it. The only problem I see is that the tests are never run in a browser environment otherwise, they will only get run in a node environment which can be quite different.

tyleryasaka:
I think we should maybe keep a standalone command/script that can start the test server, but it should not be included with `npm start`. It could be useful to run every once in a while but it doesn't seem like anyone is using it on a regular basis.

PR #488, opened 2018-09-18
Remove V00 marketplace and rename V01 to V00

tyleryasaka:
### Checklist:

### Description:

We had an extra marketplace version to test upgradability, but there has been some confusion as to which version to use.


micahalcorn:
@wanderingstan I'll let you confirm and merge since you were more involved in this discussion.

tyleryasaka:
I checked with @wanderingstan first and he said he's good with the naming (V00 vs V01). He hadn't deployed anything yet.

PR #490, opened 2018-09-19
Add marketplace features to token CLI

cuongdo:
### Checklist:

### Description:

Originally, I'd wanted to create a separate marketplace CLI, but there
is no time for that now. So, I've add 3 marketplace actions to the token
CLI.

While doing this, I also refactored common functionality out of
token/lib/token.js into token/lib/_contractHelper.js.

### DIFF
--- a/token/lib/_contractHelper.js
+++ b/token/lib/_contractHelper.js
@@ -1,0 +1,212 @@
+const Web3 = require('web3')
+
+const IMultiSigWallet = require('../../contracts/build/contracts/IMultiSigWallet.json')
+const { withRetries } = require('../../src/utils/retries.js')
+
+const { isValidOwner, validOwners } = require('./owner_whitelist.js')
+
+class ContractHelper {
+  constructor(config) {
+    this.config = config
+    this.retries = 7
+    this.validOwners = validOwners
+  }
+
+  /**
+   * Returns a web3 object with a provider for the given network ID.
+   * @params {string} networkId - Network ID.
+   * @returns {object} - Promise that resolves to contract object.
+   */
+  web3(networkId) {
+    const provider = this.config.providers[networkId]
+    return new Web3(provider)
+  }
+
+  /**
+   * Changes the owner of the given contract to the given address.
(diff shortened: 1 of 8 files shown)


token/lib/_contractHelper.js
```diff
@@ -0,0 +1,212 @@
...
+    if (!opts.gas) {
+      opts.gas = await transaction.estimateGas({ from: opts.from })
+      this.vlog('estimated gas:', opts.gas)
+    }
+
+    // Send the transaction and grab the transaction hash when it's available.
+    this.vlog('sending transaction')
+    let transactionHash
+    transaction.send(opts)
+      .on('transactionHash', (hash) => {
```
franckc: Enhancement idea: I wonder if it would be simpler to use the ["confirmation" callback](https://web3js.readthedocs.io/en/1.0/web3-eth.html#eth-sendtransaction-return). That callback includes the receipt. As a bonus it would allow us to be extra safe by waiting for N block confirmations to make sure the block indeed made it to the blockchain.

token/lib/_contractHelper.js
```diff
@@ -0,0 +1,212 @@
...
+    // around some strange interactions between web3.js and some web3 providers.
+    // For example, this issue sometimes prevents transaction receipts from
+    // being returned when simply calling
+    // `await contract.methods.MyMethod(...).send(...)`:
+    //
+    // https://github.com/INFURA/infura/issues/95
+
+    // Blocks are mined every ~15 seconds, but it sometimes takes ~40-60 seconds
+    // to get a transaction receipt from rinkeby.infura.io.
+    const maxSleep = 120000
```
franckc: What are maxSleep, totalSleep, sleepTime used for ? I did not see them referenced in the code.

token/lib/_contractHelper.js
cuongdo: For some contract calls, the `receipt` callback never fires for me using rinkeby.infura.io. I'm curious if the `confirmation` callback will. I'll try it sometime.

PR #493, opened 2018-09-19
Add missing dependancies

sparrowDom:
### Description:

Just did a fresh install of `origin-box` and was missing `elasticsearch`, `pg`, `graphql` & `apollo server` dependancies. Theese are needed so that elasticsearch indexer and apollo server can be started


DanielVF:
I think currently these are in a separate NPM package in daemons/indexing? These should definitely not be a part of origin-js's dependencies.

I usually bash into the container, `cd daemons/indexing && npm install`. This should be done by the box, rather than manualy.

sparrowDom:
@DanielVF ahhh yes you are right, my bad. Let me cancel this

PR #497, opened 2018-09-19
Clean up status logic in getOffer()

jordajm:
### Checklist:

### Description:

- Cleans up logic that assigns a status to an offer in `getOffer()` function
- Fixes mis-named listing status `resolved` to be called `ruling` to match corresponding offer status


src/adapters/marketplace/v00.js
```diff
@@ -225,7 +225,7 @@ class V00_MarkeplaceAdapter {
       } else if (event.event === 'OfferDisputed') {
         offers[event.returnValues.offerID] = { status: 'disputed', event }
       } else if (event.event === 'OfferRuling') {
-        offers[event.returnValues.offerID] = { status: 'resolved', event }
+        offers[event.returnValues.offerID] = { status: 'ruled', event }
```
wanderingstan: The grammar is a bit odd, but I'd rather stay consistent with having js status' match the event names, and have this as `ruling` rather than `ruled`.

jordajm:
Updated @wanderingstan - `"ruled"` is now `"ruling"`

src/contractInterface/marketplace/v00_adapter.js
```diff
@@ -310,11 +310,18 @@ class V00_MarkeplaceAdapter {
         ipfsHash = e.returnValues.ipfsHash
         createdAt = timestamp
         break
+      case 'OfferAccepted':
```
franckc: No need to update status in case of OfferAccepted and OfferDisputed.
Since the Offer data is still onchain at this stage (e.g. the Offer data has NOT been deleted from onchain), when rawOffer is loaded at line 291 above, it will return an object with status populated.

jordajm:
Good call @franckc - updated

PR #507, opened 2018-09-20
CLI support for mainnet and other improvements

cuongdo:
### Checklist:

### Description:

* Add support for mainnet
* Reduced verbosity of retry logging
* Cleaned up `withRetries` interface
* Improved some error messages

Together, these changes give me more confidence that the CLI will work with the testnets and mainnet. I tested the CLI with ganache-cli, Rinkeby, and (in a very limited way) mainnet.

### DIFF
--- a/token/lib/_contractHelper.js
+++ b/token/lib/_contractHelper.js
@@ -94,10 +94,11 @@ class ContractHelper {
     this.vlog('sending transaction')
     let transactionHash
     transaction.send(opts)
-      .on('transactionHash', (hash) => {
+      .once('transactionHash', (hash) => {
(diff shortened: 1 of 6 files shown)


token/lib/_contractHelper.js
```diff
@@ -94,10 +94,11 @@ class ContractHelper {
     this.vlog('sending transaction')
     let transactionHash
     transaction.send(opts)
-      .on('transactionHash', (hash) => {
+      .once('transactionHash', (hash) => {
```
franckc: wow. good catch. it is kind of confusing web3 uses both 'on' and 'once'... :(

PR #518, opened 2018-09-20
Use untranslated fixtures

sparrowDom:
### Description:

Since we store untranslated category and subcategory strings in index/ipfs ( https://github.com/OriginProtocol/origin-dapp/pull/525/files ) this also updates our fixtures. 

Connected PR: https://github.com/OriginProtocol/origin-dapp/pull/535

### DIFF
--- a/daemon/indexing/listener/listener.js
+++ b/daemon/indexing/listener/listener.js
@@ -1,7 +1,7 @@
 const fs = require('fs')
 const http = require('http')
 const urllib = require('url')
-const Origin = require('origin')
+const Origin = require('../../../dist/index')
 const Web3 = require('web3')
 
 const search = require('../lib/search.js')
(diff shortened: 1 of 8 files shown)


daemon/indexing/listener/listener.js
```diff
@@ -1,7 +1,7 @@
 const fs = require('fs')
 const http = require('http')
 const urllib = require('url')
-const Origin = require('origin')
+const Origin = require('../../../dist/index')
```
franckc: I think @tomlinton updated the origin-js container to generate a local origin npm package. so if you update origin-box and recreate your origin-js container, you should not need to change this ?

sparrowDom: Ahh yes this should not be in, was using it for local testing. Removing this change...

sparrowDom: Thanks for catching this, removed...

Issue #457, opened 2018-09-14
Contract Response Codes 


micahalcorn:
Add response codes to all require statements so that contract consumers can get insights into why contract methods are failing.

cuongdo:
I'm going to put in reason strings for our various contracts

PR #519, opened 2018-09-20
Add reasons for requires in contracts

cuongdo:
### Checklist:

### Description:

Add reason strings to requires

This makes the cause of an EVM revert much clearer to developers. I've
tried to keep the strings short but unambiguous within each function.

I've had to increase the gas limit for contract test deploys, because it
now takes ~5 million gas (compared to ~4 million previously) to deploy
the marketplace contract. This corresponds to a difference in deployment
cost of ~$1.31 USD. The small increase is more than offset by reduction
in developer debugging time. The lowest gas limit I've seen from
Rinkeby, Ropsten, and mainnet is 7 million, so we're OK there.

I also haven't touched contracts/test-alt/contracts

Takes care of #457 for Origin contract code

OpenZeppelin code remains uninstrumented until we vendor it.

### DIFF
--- a/contracts/contracts/marketplace/v00/Marketplace.sol
+++ b/contracts/contracts/marketplace/v00/Marketplace.sol
@@ -88,7 +88,7 @@ contract V00_Marketplace is Ownable {
   )
     public returns (bool)
   {
-    require(msg.sender == address(tokenAddr));
+    require(msg.sender == address(tokenAddr), "token must call");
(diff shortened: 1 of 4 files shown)


cuongdo:
cc @wanderingstan is this what you were thinking?

wanderingstan:
Sorry this is late, but YES! Thank you @cuongdo !

PR #536, opened 2018-09-24
Remove duplicate listener rules

wanderingstan:
Looks like we got a duplicate left over from when we had the v00 marketplace lying around.

EDIT: Or...do we need two entries in this array for some reason? (I've never worked in the listener codebase!)


sparrowDom:
LGTM

franckc:
Good eye ! Thanks @wanderingstan :)

PR #542, opened 2018-09-25
Token whitelist simplification & testnet whitelist

cuongdo:

### Description:

I followed @nick's suggestion to combine the token allowed sender and recipient lists. That should make the code easier to follow and leave less room for bugs or misconfigurations of the whitelist.

I also added a Truffle migration to set up the whitelist for the testnets. We should use the whitelist there before relying on it for mainnet.


cuongdo:
cc @franckc

PR #544, opened 2018-09-25
Affiliate whitelist

nick:

### Checklist:

### Description:

- Adds new methods `addAffiliate` and `removeAffiliate` to Marketplace contract.
- Using addAffiliate with the marketplace contract address effectively disabled the whitelist such that any affiliate may be used.
- To keep consistency, you can also add an IPFS hash when adding or removing an affiliate which could contain more information about the new affiliate or why an affiliate was removed.

### DIFF
--- a/contracts/contracts/marketplace/v00/Marketplace.sol
+++ b/contracts/contracts/marketplace/v00/Marketplace.sol
@@ -35,6 +35,8 @@ contract V00_Marketplace is Ownable {
     event OfferFinalized   (address indexed party, uint indexed listingID, uint indexed offerID, bytes32 ipfsHash);
     event OfferData        (address indexed party, uint indexed listingID, uint indexed offerID, bytes32 ipfsHash);
     event MarketplaceData  (address indexed party, bytes32 ipfsHash);
+    event AffiliateAdded   (address indexed party, bytes32 ipfsHash);
+    event AffiliateRemoved (address indexed party, bytes32 ipfsHash);
 
     struct Listing {
         address seller;     // Seller wallet / identity contract / other contract
@@ -56,6 +58,7 @@ contract V00_Marketplace is Ownable {
 
     Listing[] public listings;
     mapping(uint => Offer[]) public offers; // listingID => Offers
+    mapping(address => bool) public allowedAffiliates;
 
     ERC20 public tokenAddr; // Origin Token address
 
@@ -182,6 +185,11 @@ contract V00_Marketplace is Ownable {
         public
         payable
     {
+        require(
+            _affiliate == 0x0 || allowedAffiliates[address(this)] || allowedAffiliates[_affiliate],
+            "Affiliate not allowed"
+        );
+
         offers[listingID].push(Offer({
             status: 1,
             buyer: msg.sender,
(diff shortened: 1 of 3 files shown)


contracts/test-alt/Marketplace.js
```diff
@@ -643,4 +643,60 @@ describe('Marketplace.sol', async function() {
       }
     })
   })
+
+  describe('Affiliate whitelist', function() {
```
cuongdo: could you add a test to ensure that only the owner can add or remove affiliates? I know it's obvious from the modifier, but it's extra piece of mind

contracts/contracts/marketplace/v00/Marketplace.sol
```diff
@@ -182,6 +185,11 @@ contract V00_Marketplace is Ownable {
         public
         payable
     {
+        require(
+            _affiliate == 0x0 || allowedAffiliates[address(this)] || allowedAffiliates[_affiliate],
```
cuongdo: should we allow `0x0`?

cuongdo: `address(this)` isn't obvious for "whitelist all". maybe define a constant above (`ALL_AFFILIATES`?) and use it here?

contracts/test-alt/_marketplaceHelpers.js
```diff
@@ -47,7 +47,7 @@ export default function({
     }
   }
 
-  async function makeOffer({ withdraw, listingID = 0 }) {
+  async function makeOffer({ withdraw, listingID = 0, affiliate = Buyer }) {
```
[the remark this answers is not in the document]
cuongdo: I realize that was its value before, but I still don't understand the reasoning behind the value

contracts/test-alt/Marketplace.js
cuongdo: Actually, I just noticed that you already have tests for this.

contracts/contracts/marketplace/v00/Marketplace.sol
cuongdo: I've pushed a change that adds the constant. It shouldn't use any gas, because constants don't get storage slots.

contracts/test-alt/_marketplaceHelpers.js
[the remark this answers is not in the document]
nick: Initially it was so that the buyer can earn their own commission as a kind of 'reward points'. We could also make another account called 'Affiliate' and use that instead, to make the tests clearer

contracts/contracts/marketplace/v00/Marketplace.sol
nick: Regarding the 0x0, we could remove the check and instead add 0x0 to the whitelist in the constructor. That would give us control over whether no affiliate should be allowed

franckc:
Sorry for being late to this conversation...

If we allow affiliate to be 0x0, won't the commission get stuck in escrow forever ?

cuongdo:
@franckc that's exactly what's happening right now on master. we need to have an affiliate account on all networks and pass that through when calling `makeOffer`. I'm adding a contract-level safeguard so that we don't allow a non-zero commission if affiliate is 0x0

PR #545, opened 2018-09-25
Misc faucet fixes

franckc:

### Checklist:

### Description:
A few faucet fixes:
 - Do not show new balance on faucet result page since in some cases it was not accurate. Likely due to infura nodes load balancing/out of sync behavior (e.g. token credit gets processed by an Infura node, but subsequent balance read call goes to a different node which is out of sync and therefore returns old balance).
 - Tighten up rate limiting to max 1 req per 5 sec and apply it only to /tokens route.

### DIFF
--- a/token/faucet/app.js
+++ b/token/faucet/app.js
@@ -8,7 +8,7 @@ const Token = require('../lib/token.js')
 const DEFAULT_SERVER_PORT = 5000
 const DEFAULT_NETWORK_ID = '999' // Local blockchain.
 
-// Credit 100 token units per request.
+// Credit 100 tokens per request.
 const NUM_TOKENS = 100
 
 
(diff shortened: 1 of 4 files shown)


token/faucet/app.js
```diff
@@ -19,22 +19,26 @@ function runApp(config) {
 
   // Configure rate limiting. Allow at most 1 request per IP every 5 sec.
   const opts = {
-    points: 90,   // Point budget.
+    points: 1,   // Point budget.
     duration: 5, // Reset points consumption every 5 sec.
```
cuongdo: If I'm reading this correctly, a single IP can get ~1200 OGN per minute. Because the max boost is 100 OGN, that seems a little excessive. How about 100 or 200 OGN per minute? (if this is scripted, someone would still be able to get 8,640,000 OGN per day, which is a lot)

franckc: Yes you are correct. That's a bit high. Adjusted to allow 100 OGN per 60 sec.

PR #546, opened 2018-09-25
Sparrow dom/tranaction hash

sparrowDom:
### Checklist:

### Description:

**Connected PR: ** https://github.com/OriginProtocol/origin-dapp/pull/577

### DIFF
--- a/src/resources/users.js
+++ b/src/resources/users.js
@@ -5,8 +5,8 @@ class Users {
     this.resolver = new UsersResolver({ contractService, ipfsService })
   }
 
-  async set({ profile, attestations = [], options = {} }) {
-    return this.resolver.set({ profile, attestations, options })
+  async set({ profile, transactionHashCallback, attestations = []}) {
+    return this.resolver.set({ profile, transactionHashCallback, attestations })
   }
 
   async get(address) {
--- a/src/resources/users.js
+++ b/src/resources/users.js
@@ -5,8 +5,8 @@ class Users {
     this.resolver = new UsersResolver({ contractService, ipfsService })
   }
 
-  async set({ profile, attestations = [] }) {
-    return this.resolver.set({ profile, attestations })
+  async set({ profile, attestations = [], options = {} }) {
+    return this.resolver.set({ profile, attestations, options })
   }
 
   async get(address) {
--- a/src/resources/users.js
+++ b/src/resources/users.js
@@ -5,6 +5,10 @@ class Users {
     this.resolver = new UsersResolver({ contractService, ipfsService })
   }
 
+  /* possible options values: 
+   * - confirmationCallback(confirmationCount, transactionReceipt) -> called repeatedly after a transaction is mined
+   * - transactionHashCallback(hash) -> called immediately when the transaction hash is received
+   */
   async set({ profile, attestations = [], options = {}}) {
     return this.resolver.set({ profile, attestations, options })
   }
--- a/src/resources/users.js
+++ b/src/resources/users.js
@@ -5,8 +5,8 @@ class Users {
     this.resolver = new UsersResolver({ contractService, ipfsService })
   }
 
-  async set({ profile, transactionHashCallback, attestations = []}) {
-    return this.resolver.set({ profile, transactionHashCallback, attestations })
+  async set({ profile, attestations = [], options = {}}) {
+    return this.resolver.set({ profile, attestations, options })
   }
 
   async get(address) {
--- a/src/services/contract-service.js
+++ b/src/services/contract-service.js
@@ -214,7 +214,7 @@ class ContractService {
     contractName,
     functionName,
     args = [],
-    { contractAddress, from, gas, value, confirmationCallback, additionalGas = 0 } = {}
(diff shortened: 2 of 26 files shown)


franckc:
Thanks @sparrowDom for working on this !
I can review once the discussion regarding approach to take on DApp side is settled a bit more... regardinhttps://github.com/OriginProtocol/origin-dapp/pull/577

sparrowDom:
@franckc yup I guess this is currently more in a suggestion/proof of concept stage.

sparrowDom:
@tyleryasaka can you please take another look. Have added a more general `options` approach that could be used in unification api step that we will do some time later.

src/services/contract-service.js
```diff
@@ -205,16 +205,53 @@ class ContractService {
...
+      setTimeout(() => {
+        this.checkForDeploymentCompletion(hash, confirmationCallback)
+      }, 1500)
+    } else {
+      const currentBlockNumber = await this.web3.eth.getBlockNumber()
+      const confirmations = currentBlockNumber - transactionInfo.blockNumber
+      confirmationCallback(confirmations, {
+        transactionHash: transactionInfo.hash
+      })
+      // do checks untill 5 block confirmations
```
franckc: typo: "until"
nit: let's define 5 as a constant. Like NUM_CONFIRMATIONS

src/resources/users.js
```diff
@@ -5,8 +5,8 @@ class Users {
     this.resolver = new UsersResolver({ contractService, ipfsService })
   }
 
-  async set({ profile, attestations = [] }) {
-    return this.resolver.set({ profile, attestations })
+  async set({ profile, attestations = [], options = {}}) {
```
franckc: nit: can we document as a comment what options are available ?

sparrowDom:
@franckc thanks for reminding me about tests. Inline comments are addressed and a test added.

test/resource_users.test.js
```diff
@@ -220,6 +222,30 @@ describe('User Resource', function() {
...
+          },
+          confirmationCallback: (confirmationCount, transactionReceipt) => {
+            confCount = confirmationCount
+            trReceipt = transactionReceipt
+          }
+        }
+      })
+
+      expect(transactionHash).to.startsWith('0x')
+      expect(confCount).is.a('number')
```
franckc: Isn't there a possible race condition here since it is not guaranteed the confirmationCallback will get called before this line is executed ?

One possible way to test callbacks I think is to put the test code within the callback itself. See for ex: https://gist.github.com/eswak/c54989f05c6f2ce4be00
Maybe something along the lines of
```
confirmationCallback: (confirmationCount, transactionReceipt) => {
  expect(confirmationCount).is.a('number')
  expect(trReceipt).is.a('object')
  if (confirmationCount == NUMBER_CONFIRMATIONS_TO_REPORT) {
    // this test is done, go to the next one
    done()
  }
```

src/services/contract-service.js
```diff
@@ -13,6 +13,7 @@ import bs58 from 'bs58'
 import Web3 from 'web3'
 
 const emptyAddress = '0x0000000000000000000000000000000000000000'
+const NUMBER_CONFIRMATIONS_TO_REPORT = 20
```
franckc: Is 20 a bit too high ? I thought typically 6 to 10 confirmations is good ? Maybe we pick 10 ?

src/services/contract-service.js
```diff
@@ -239,6 +277,7 @@ class ContractService {
         .send(opts)
         .on('receipt', resolve)
         .on('confirmation', confirmationCallback)
```
franckc: Won't we have same issue here mentioned in the deploy method that confirmationCallback does not get triggered ? Are we relying on this callback anywhere in our DApp or origin-js codebase ? If yes should we fix it, if no can we at least add a comment to warn caller that this is currently not working ?

sparrowDom: Interestingly when doing an Ethereum transaction confirmation callback is working, but when deploying a contract it is not. That is why the workaround is implemented only in the `deploy` function a few lines above.

test/resource_users.test.js
sparrowDom: Ahh yes thanks @franckc . The approach with a done() callback is much better.

sparrowDom:
@franckc responded to comments and improved the test as you have proposed. 

And please don't hold back on the Pull Request comments. Each comment results either in clearing up confusion, improving the code base and/or improving coding skills. For that reason I prefer as many comments as possible.

src/services/contract-service.js
franckc: Ok. Sg. Let's keep 20 for now then. Thanks for the details :)

src/services/contract-service.js
franckc: Got it. Thanks for being thorough.

src/services/contract-service.js
micahalcorn: My reason for 1️⃣:two:: https://ethereum.stackexchange.com/a/3009/42177

sparrowDom: Thanks for explanation @micahalcorn

Issue #541, opened 2018-09-24
Validate addresses of arbitrator, affiliate, messaging accounts 


franckc:
This is part of hardening origin-js.

At the origin-js layer, we should throw an error in the following scenarios:
 - ~set messaging account to anything else than official Origin messaging account~
 - create listing that uses non official Origin arbitrator or affiliate account
 - read a listing where arbitrator or affiliate is not official Origin account

franckc:
@tyleryasaka we discussed the above during our security brainstorm this afternoon. Would you be ok adding some validation for this to origin-js ? Thank you sir !

micahalcorn:
There is no concept of the messaging resource being aware or caring about who is generating what message. So I think it's the DApp's responsibility to indicate whether or not a messaging user is a trusted Origin account. If I'm right, then I'm not sure why Origin.js needs that value. I'll update @franck's original comment.

micahalcorn:
I also think that @nick intended for the arbitrator on the listing to be the seller vs the arbitrator on the offer being the [Origin] centralized arbitrator in question.

Regardless, there is still a question of whether we hard-code the official Origin arbitrator and affiliate address in Origin.js for the time being or just let the DApp pass them in (from env vars) and go ahead and allow other DApps to potentially supply different values. cc @wanderingstan

wanderingstan:
As for which layer has the hard-coded address, I say we go with whatever is the easiest for now. Either approach will need adjustments once we have other partners/developers building on our code, and we could use a few more thought cycles on where the right place is. (I idealistically lean toward it being the DApp, but recognize that security may tilt us towards origin.js package.)

tyleryasaka:
Discussed with @franckc , these were the takeaways:
- @micahalcorn you are correct, the arbitrator on the listing is currently being set as the seller and this appears to be intentional. The arbitrator on the offer is the entity that will resolve disputes between the buyer and seller on an offer.
- We will pass in arbitrator and affiliate for offers as config options from the Dapp to origin-js. This allows them to be env vars in the Dapp that are passed in to origin-js at runtime, vs having to check in those addresses into source code.
[...]

PR #548, opened 2018-09-25
Nick/affiliate whitelist2

nick:

### Checklist:

### Description:

- Adds new methods `addAffiliate` and `removeAffiliate` to Marketplace contract.
- Using addAffiliate with the marketplace contract address effectively disabled the whitelist such that any affiliate may be used.
- To keep consistency, you can also add an IPFS hash when adding or removing an affiliate which could contain more information about the new affiliate or why an affiliate was removed.


cuongdo:
Also, this cannot merge until @tyleryasaka fixes #541. Otherwise, this will cause offer creation to fail because affiliate is not currently set by the dapp.

tyleryasaka:
Yeah if you don't mind, for https://github.com/OriginProtocol/origin-js/pull/549 to be merged

tyleryasaka:
Also @nick maybe look at my PR #549 ? Not sure if we need to do affiliate whitelisting at the contract level if we're doing that validation in origin-js (which is what my PR adds). best to keep complexity out of the contract, right?

cuongdo:
pushed test fixes. apparently, our tests use a separately deployed marketplace contract. 🙏

cuongdo:
@franckc @nick I just pushed an updated migration that whitelists the affiliate for each network, except for mainnet (I don't have that address yet). Please take a look

cuongdo:
I was able to fully test a transaction with commission with the following additions to my DApps' `.env` file:

```
AFFILIATE_ACCOUNT=0x627306090abab3a6e1400e9345bc60c78a8bef57
ARBITRATOR_ACCOUNT=0x821aea9a577a9b44299b9c15c88cf3087f3b5544
```

The commission ends up in the AFFILIATE_ACCOUNT as expected, which is great.

So, I believe this is good to merge after @franckc gets both account variables in the various environments.

franckc:
Sent out this [PR](https://github.com/OriginProtocol/origin-box/pull/89) to set the addresses on local, dev and staging environments.

PR #549, opened 2018-09-25
Validate arbitrator & affiliate

### LINKED ISSUE #541: Validate addresses of arbitrator, affiliate, messaging accounts

tyleryasaka:
### Checklist:

### Description:

`arbitrator` and `affiliate` can now be passed in as config options at initialization of origin-js. Offers are automatically populated with and validated against these values.

Resolves #541

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -26,6 +26,8 @@ class Origin {
     ipfsGatewayProtocol = defaultIpfsGatewayProtocol,
     attestationServerUrl = defaultAttestationServerUrl,
     discoveryServerUrl = defaultDiscoveryServerUrl,
+    affiliate,
+    arbitrator,
     contractAddresses,
     web3,
     ipfsCreator,
@@ -52,6 +54,8 @@ class Origin {
     this.marketplace = new Marketplace({
       contractService: this.contractService,
       ipfsService: this.ipfsService,
+      affiliate,
+      arbitrator,
       store
     })
 
--- a/src/resources/marketplace.js
+++ b/src/resources/marketplace.js
@@ -18,9 +18,11 @@ import {
 import MarketplaceResolver from '../contractInterface/marketplace/resolver'
 
 class Marketplace {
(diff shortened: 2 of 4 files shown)


micahalcorn:
I understand the reason for validating offers when they are retrieved rather than created, but shouldn't we at least require the presence of an affiliate address when the offer is created? If a DApp forgets to pass the affiliate param, it will seemingly "work" but the offer will never be found. In that case, the developer isn't a bad actor - just a naive actor.

tyleryasaka:
@micahalcorn there's nothing to validate because we don't even accept affiliate address as a parameter. It's automatically populated, every time! 😄

tyleryasaka:
The only way you can create an offer without an affiliate is to manually call the smart contracts yourself.

micahalcorn:
@tyleryasaka fair point, but it will still fail if there is no environment variable set. We can worry about it later.

tyleryasaka:
The only place the Dapp could mess up is when initializing origin-js. I suppose if you forgot the affiliate in the config that could be problematic.

A possible feature that could be added in a separate PR might throw an error if origin-js is initialized without `arbitrator` or `affiliate` options.

micahalcorn:
Yeah, I like that.

tyleryasaka:
For that matter, might want to also validate the presence of other necessary config options.

tyleryasaka:
@micahalcorn I filed an issue: https://github.com/OriginProtocol/origin-js/issues/551

micahalcorn:
Thanks @tyleryasaka 🙏 One other issue with this is that it produces errors when someone fetches all notifications. It's an unfortunate experience when someone requests notifications and they end up getting errors from offers that they didn't consciously ask for.

PR #550, opened 2018-09-25
Add support for private key provider to truffle config.

franckc:

### Checklist:

### Description:
Added support for private key provider to the truffle config.
This will allow us to use a throwaway hot wallet for doing contracts deployment, by setting the PRIVATE_KEY env variable before running the truffle migrate command.

### DIFF
--- a/contracts/truffle.js
+++ b/contracts/truffle.js
@@ -1,7 +1,10 @@
 const HDWalletProvider = require("truffle-hdwallet-provider")
+const PrivateKeyProvider = require('truffle-privatekey-provider')
+
 
 // How many addresses in wallet should we unlock?
 // (For deploying test data, we use other addresses as buyers and sellers)
+// Note: This is not used for Mainnet - only for Testnet and local deployment.
 const numAddressesToUnlock = 4
 
 // Local setup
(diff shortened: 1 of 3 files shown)


contracts/truffle.js
```diff
@@ -1,7 +1,10 @@
 const HDWalletProvider = require("truffle-hdwallet-provider")
+const PrivateKeyProvider = require('truffle-privatekey-provider')
```
wanderingstan: Not a huge deal, but it looks like we may not need this package. This note appears on a similar functioning package [truffle-hdwallet-provider-privkey](https://www.npmjs.com/package/truffle-hdwallet-provider-privkey)
> If you are using Web3 1.0.0, you do not need this package! There is a function to get an account object from a private key. The reason I made this package was because this function doesn't exist in the legacy Web3 library.

franckc:
Ah interesting. Yes, this npm package does not implement a lot:
https://github.com/nosuchip/truffle-privatekey-provider/blob/master/index.js#L9

Though still it seems convenient to not rewrite that code ?

wanderingstan:
Agreed, .good not to rewrite that code, and it gives us similar interface as `truffle-hdwallet-provider`.

There's of course always a risk in having another dependency, but I feel good on this one.

PR #556, opened 2018-09-26
Fix getOffer() arbitrator and affiliate checking

cuongdo:
### Checklist:

### Description:

Fix getOffer() arbitrator and affiliate checking

Sometimes, some characters in addresses are upper-cased. Not sure why,
but the net result is that addresses must be compared case
insensitively.  I observed this locally, and it prevented in-progress
offers from showing up on the buyer or seller side.

### DIFF
--- a/src/resources/marketplace.js
+++ b/src/resources/marketplace.js
@@ -133,11 +133,11 @@ class Marketplace {
         throw new Error('Invalid offer: insufficient commission amount for listing')
       }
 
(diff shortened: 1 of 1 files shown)


cuongdo:
here's why addresses are sometimes mixed case: https://forum.ethereum.org/discussion/9220/eth-address-upper-and-lower-characters-does-not-matter

PR #563, opened 2018-09-27
Mainnet truffle migrations

cuongdo:
### Checklist:

### Description:

This should make us ready for mainnet deployment of the tokens. Added in this PR are:

* New migration to transfer OGN, marketplace contract ownership, and token contract ownership to mainnet multi-sig
* New mainnet verification migration: this doesn't migrate anything but is a useful set of automated, gas-free tests
* A new `SIMULATE_MAINNET` environment variable to try truffle migrations in "mainnet mode" on your local blockchain
  * e.g. `SIMULATE_MAINNET=1 npx truffle migrate --reset`
* Fix for a buggy error message encountered in a Rinkeby deployment
* Add multi-sig wallets to token CLI owner whitelist


cuongdo:
cc @joshfraser 

Please pay extra attention to the mainnet addresses. Thanks!

contracts/migrations/7_create_token_whitelist.js
```diff
@@ -5,23 +5,12 @@ const V00_Marketplace = artifacts.require('./V00_Marketplace.sol')
...
-  })
+  if (network !== 'development' || process.env['SIMULATE_MAINNET']) {
+    return createTokenWhitelist(network)
+  }
 }
 
-async function createTokenWhitelistForTestnets() {
+async function createTokenWhitelist(network) {
   const token = await OriginToken.deployed()
   const tokenOwner = await token.owner()
```
franckc: We won't have a faucet on Mainnet. So should this be only executed if network is not mainnet ?

contracts/migrations/7_create_token_whitelist.js
```diff
@@ -36,6 +25,17 @@ async function createTokenWhitelistForTestnets() {
   await token.addAllowedTransactor(marketplace.address, { from: tokenOwner })
   console.log(`Added marketplace ${marketplace.address} to whitelist`)
 
+  if (network === 'mainnet' || process.env['SIMULATE_MAINNET']) {
+    const addresses = [
```
franckc: nit: maybe add a comment that this is for adding the affiliate and ERC20 multisig wallet.

contracts/migrations/9_transfer_ownership_and_tokens_for_mainnet.js
```diff
@@ -0,0 +1,47 @@
+const OriginToken = artifacts.require('./token/OriginToken.sol')
+const V00_Marketplace = artifacts.require('./V00_Marketplace.sol')
+
+const tokenMultiSig = '0xe011fa2a6df98c69383457d87a056ed0103aa352'
+const marketplaceMultiSig = '0x8a1a4f77f9f0eb35fb9930696038be6220986c1b'
+
+module.exports = function(deployer, network) {
+  return deployer.then(() => {
+    if (network === 'mainnet' || process.env['SIMULATE_MAINNET']) {
+      return transferTokensAndContractsToMultiSig()
```
franckc: do we need an await here since transferTokensAndContractsToMultiSig is an async function otherwise truffle could mark this migration step as successful before it finishes to run ?

contracts/migrations/9_transfer_ownership_and_tokens_for_mainnet.js
```diff
@@ -0,0 +1,47 @@
+const OriginToken = artifacts.require('./token/OriginToken.sol')
+const V00_Marketplace = artifacts.require('./V00_Marketplace.sol')
+
+const tokenMultiSig = '0xe011fa2a6df98c69383457d87a056ed0103aa352'
```
franckc: Not needed in this PR but maybe add a TODO for us to put all these addresses in a common "addresses.js" file so that they can then be imported by all the migrations scripts. Otherwise I'm worried in the future we may update an address in one migration script but forget to update in the others ?

contracts/migrations/7_create_token_whitelist.js
cuongdo: `transferTokensAndContractsToMultiSig()` in migration 9 removes the token deployer from the whitelist for mainnet, so I think this is fine

contracts/migrations/7_create_token_whitelist.js
cuongdo: done

contracts/migrations/9_transfer_ownership_and_tokens_for_mainnet.js
cuongdo: added TODO

contracts/migrations/9_transfer_ownership_and_tokens_for_mainnet.js
cuongdo: truffle migrate runs the migrations [using 
 async.eachSeries](https://github.com/trufflesuite/truffle/blob/develop/packages/truffle-migrate/index.js#L194), so it'll wait for the promise to resolve

contracts/migrations/7_create_token_whitelist.js
cuongdo: the deployer also needs to transfer all its OGN to the multi-sig wallet, so it needs to be whitelisted temporarily

PR #569, opened 2018-09-28
Fix for Rinkeby "nonce too low" issue

cuongdo:
### Checklist:

### Description:

I've put in a NonceTrackerSubprovider for every network except
development (I'm not sure how to do it there).

Tested with:
- Rinkeby
- Ropsten

NOT tested with mainnet due to insufficient ETH.

### DIFF
--- a/contracts/truffle.js
+++ b/contracts/truffle.js
@@ -1,5 +1,6 @@
 const HDWalletProvider = require("truffle-hdwallet-provider")
 const PrivateKeyProvider = require('truffle-privatekey-provider')
+const NonceTrackerSubprovider = require("web3-provider-engine/subproviders/nonce-tracker")
 
 
 // How many addresses in wallet should we unlock?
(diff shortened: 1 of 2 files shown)


contracts/truffle.js
```diff
@@ -1,5 +1,6 @@
 const HDWalletProvider = require("truffle-hdwallet-provider")
 const PrivateKeyProvider = require('truffle-privatekey-provider')
+const NonceTrackerSubprovider = require("web3-provider-engine/subproviders/nonce-tracker")
```
franckc: Do we need to add web3-provider-engine as a dev dependency in package.json ? I think I had to install it before being able to run the truffle script with this change.

cuongdo: Done.

Though `web3-provider-engine` is already a transitive dependency through `truffle-hdwallet-provider`, you're right that it's still good to have an explicit dependency.

PR #574, opened 2018-09-28
 🎧 Listener - HTTPS webhook support

DanielVF:
Allows the listener to post to HTTP or HTTPS webhook urls. Previously we only supported HTTP.

Probably a lower priority at this time, and can go in after the mainnet beta next week.

### DIFF
--- a/daemon/indexing/listener/listener.js
+++ b/daemon/indexing/listener/listener.js
@@ -1,5 +1,6 @@
 const fs = require('fs')
 const http = require('http')
+const https = require('https')
 const urllib = require('url')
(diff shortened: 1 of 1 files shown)


daemon/indexing/listener/listener.js
```diff
@@ -1,5 +1,6 @@
 const fs = require('fs')
 const http = require('http')
+const https = require('https')
```
franckc: do we need to update package.json ?

DanielVF: Nope! These are built into node.

PR #590, opened 2018-10-02
Tyleryasaka/clean build script

### LINKED ISSUE #491: Optimise build script and allow builds when NODE_ENV is set to production
It is not possible to do a production build without installing all of the devDependencies from the `package.json`. This is very slow, and since we have to build `origin-js` for other parts of the stack it makes updating deployments very slow also.

The build script [here](https://github.com/OriginProtocol/origin-js/blob/master/scripts/build.js) imports a lot of things we don't need for a production build (`ganache`, `ipfs`, `chalk`, `node-watch` look like examples of this, there might be more). If we could avoid installing all of this stuff every time we do a production build the build time would be cut way down.

The `build.js` script at the moment is more like a mini dev environment rather than a build script. I think in the long term we should move some of the services it creates (`ganache`, `ipfs`) to `origin-box`. In the short term, perhaps a separate script to just do builds of `origin-js`?

tyleryasaka:
### Checklist:

### Description:

See linked issue below. @tomlinton does this take care of the issue for you?

Resolves #491


tomlinton:
@tyleryasaka looks great thanks! I'll test it out with some container builds.

@franckc If we are going to be doing much more deploying it'd be good to look at merging this so I can make staging match prod with a production build of origin-js. That will save us from issues like the one yesterday where the environments differ and we only discover the problem after we publish the package to npm. I promise to test it thoroughly so it doesn't break anything. :grin:

franckc:
Let me submit a PR for this cleanup on monorepo...

franckc:
Moved to monorepo: https://github.com/OriginProtocol/origin/pull/758

Closing this PR.

PR #596, opened 2018-10-03
create a fix for transaction hash

sparrowDom:
### Description:

Fixes a bug in contract service


sparrowDom:
Good news just testing this on the **Mainnet**, and while we do show an error in our modal, the confirmation workaround with this PR fix works on Metamask 4.12.0. See the transactions dropdown in the backgroud. 
screenshot 2018-10-03 23 04 14

I think this is good to merge @franckc!

sparrowDom:
Oh yeah and as expected Metamask still fumbles and throws its inner errors

franckc:
Thanks for testing @sparrowDom - going to merge now....

PR #599, opened 2018-10-03
Confirmation polling on transactions

sparrowDom:

### Checklist:

- ~[ ] Update any relevant READMEs and [docs](https://github.com/OriginProtocol/docs)~

**TODO:**
- in case we decide to merge this, tests still need to be written

### Description:
Metamask 4.12.0 introduced some breaking changes where `confirmation` and `receipt` events are not always triggered. And sometimes an error is thrown in spite of blockchain transaction succeeding. 

The solution this PR proposes: 
- for easier handling refactor so that contract function calls and contract deploys are handled by the same function
- keep `confirmation` and `receipt` events present and stop fallback solution in case those events are triggered
- when submitting transaction start a fallback pooling function that: 
  - continually pools for transaction receipt. If it finds it and `confirmation` event has not been triggered. Manually call confirmation callback
  - if transaction receipt has not been returned as promise resolution, mock it and resolve promise with it. 
- when we receive an error from Metamask that we know is not breaking don't reject the promise. Let fallback function handle resolution. 

**Related issues:**
- https://github.com/OriginProtocol/origin-dapp/issues/639
- https://github.com/OriginProtocol/origin-dapp/issues/673

### DIFF
--- a/src/services/contract-service.js
+++ b/src/services/contract-service.js
@@ -193,60 +193,81 @@ class ContractService {
     return withLibraryAddresses
   }
 
-  async deploy(contract, args, options, { confirmationCallback, transactionHashCallback } = {} ) {
-    const bytecode = await this.getBytecode(contract)
-    const deployed = await this.deployed(contract)
-    const txReceipt = await new Promise((resolve, reject) => {
-      deployed
-        .deploy({
-          data: bytecode,
-          arguments: args
-        })
-        .send(options)
-        .on('receipt', receipt => {
-          resolve(receipt)
-        })
-        //.on('confirmation', confirmationCallback)
-        //.on('transactionHash', transactionHashCallback)
-        // Workaround for "confirmationCallback" not being triggered in some versions of Metamask
-        .on('transactionHash', (hash) => {
-          if (transactionHashCallback)
-            transactionHashCallback(hash)
-          if (confirmationCallback)
-            this.checkForDeploymentCompletion(hash, confirmationCallback)
-        })
-        .on('error', err => reject(err))
-    })
-    return txReceipt
-  }
-
-  /* confirmation callback does not get triggered in current version of web3 version:1.0.0-beta.34
-   * so this function perpetually (until 20 confirmations) checks for presence of deployed contract.
+  /* Confirmation callback does not get triggered in some versions of Metamask so this
+   * function perpetually (until NUMBER_CONFIRMATIONS_TO_REPORT confirmations) checks
+   * for transaction confirmation.
    *
-   * This could also be a problem in Ethereum node: https://github.com/ethereum/web3.js/issues/1255
+   * This function is a safety net just in case send method's `on.('confirmation')` does
+   * not function (e.g. in Metamask 4.12.0 it is broken). But if we detect that the mentioned
+   * function is not broken we cancel the safety net immediately.
    */
-  async checkForDeploymentCompletion(hash, confirmationCallback) {
+  async checkForTransactionCompletion(hash, confirmationCallback, transactionStatus) {
+    // on.('confirmation') works cancel this fallback functionality
+    if (transactionStatus.onConfirmationTriggered)
+      return
+
     const transactionInfo = await this.web3.eth.getTransaction(hash)
 
     // transaction not mined
     if (transactionInfo === null || transactionInfo.blockNumber === null){
       setTimeout(() => {
-        this.checkForDeploymentCompletion(hash, confirmationCallback)
+        this.checkForTransactionCompletion(hash, confirmationCallback, transactionStatus)
       }, 1500)
     } else {
       const currentBlockNumber = await this.web3.eth.getBlockNumber()
-      // Math.max to prevent the -1 confirmation on Rinkeby
+      // Math.max to prevent the -1 confirmation on Rinkeby. 
       const confirmations = Math.max(0, currentBlockNumber - transactionInfo.blockNumber)
       confirmationCallback(confirmations, {
         transactionHash: transactionInfo.hash
       })
-      // do checks until NUMBER_CONFIRMATIONS_TO_REPORT block confirmations
+      
       if (confirmations < NUMBER_CONFIRMATIONS_TO_REPORT) {
         setTimeout(() => {
-          this.checkForDeploymentCompletion(hash, confirmationCallback)
+          this.checkForTransactionCompletion(hash, confirmationCallback, transactionStatus)
         }, 1500)
       }
     }
+  }
+
+  // Unify the way contract's send methods calls are handled
+  handleTransactionCallbacks(sendCallback, resolveCallback, rejectCallback, confirmationCallback, transactionHashCallback) {
+    const transactionStatus = { onConfirmationTriggered: false }
+    sendCallback
+      .on('receipt', resolveCallback)
+      .on('confirmation', (confirmationNumber, receipt) => {
+        transactionStatus.onConfirmationTriggered = true
+        if (confirmationCallback)
+          confirmationCallback(confirmationNumber, receipt)
+      })
+      .on('transactionHash', (hash) => {
+        if (transactionHashCallback)
+          transactionHashCallback(hash)
+        if (confirmationCallback)
+          this.checkForTransactionCompletion(hash, confirmationCallback, transactionStatus)
+      })
+      .on('error', rejectCallback)
+  }
+
+  async deploy(contract, args, options, { confirmationCallback, transactionHashCallback } = {} ) {
+    const bytecode = await this.getBytecode(contract)
+    const deployed = await this.deployed(contract)
+    const txReceipt = await new Promise((resolve, reject) => {
+      let sendCallback = deployed
+        .deploy({
+          data: bytecode,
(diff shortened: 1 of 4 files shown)


src/services/contract-service.js
```diff
@@ -276,8 +277,15 @@ class ContractService {
...
-        .on('confirmation', confirmationCallback)
-        .on('transactionHash', transactionHashCallback)
+        //.on('confirmation', confirmationCallback)
+        //.on('transactionHash', transactionHashCallback)
+        // Workaround for "confirmationCallback" not being triggered in some versions of Metamask
+        .on('transactionHash', (hash) => {
+          if (transactionHashCallback)
+            transactionHashCallback(hash)
+          if (confirmationCallback)
+            this.checkForDeploymentCompletion(hash, confirmationCallback)
```
franckc: nit: since now it is not only for deploy, can we rename checkForDeploymentCompletion to something more generic such as confirmationWatcher or something equivalent that does not have "deployment" in its name ?

src/services/contract-service.js
```diff
@@ -229,13 +229,14 @@ class ContractService {
...
     // transaction not mined
-    if (transactionInfo.blockNumber === null){
+    if (transactionInfo === null || transactionInfo.blockNumber === null){
       setTimeout(() => {
         this.checkForDeploymentCompletion(hash, confirmationCallback)
       }, 1500)
     } else {
       const currentBlockNumber = await this.web3.eth.getBlockNumber()
-      const confirmations = currentBlockNumber - transactionInfo.blockNumber
+      // Math.max to prevent the -1 confirmation on Rinkeby
```
franckc: So weird ! Why does this happen on Rinkeby only and not Mainnet ? Could be related to Infura load balancing calls across multiple nodes that may not be in sync ?

src/services/contract-service.js
sparrowDom: Yup thanks will do.

src/services/contract-service.js
sparrowDom: Yeah no idea why transactionInfo.blockNumber can be for 1 value bigger than currentBlockNumber. 

Interestingly on localhost the current block number doesn't even increase if the user is not making additional transactions. So I guess there are nuances between different chains. Or maybe metamask updates `currentBlockNumber` less frequently locally.

sparrowDom:
@micahalcorn the logic was to get some safe margin over the DApps setting of 12 confirmations that are needed for transaction to be confirmed. Perhaps the margin makes no sense and we can set it to 12 here also?

micahalcorn:
@sparrowDom I would probably assume 24 since [thats how many web3 supplies](https://web3js.readthedocs.io/en/1.0/web3-eth-contract.html?highlight=%22confirmation%22#id22).

src/services/contract-service.js
```diff
@@ -276,8 +277,15 @@ class ContractService {
       method
         .send(opts)
         .on('receipt', resolve)
-        .on('confirmation', confirmationCallback)
-        .on('transactionHash', transactionHashCallback)
+        //.on('confirmation', confirmationCallback)
```
micahalcorn: Just curious: we can't subscribe to this at all? It's what results in an exception?

src/services/contract-service.js
wanderingstan: also to nit: 
- Could we have the name including "polling", since that's the distinguishing feature vs the callback method?
- On 284, 286: We've been avoiding if clauses without `{}` brackets, and I think `npm run format` enforces this. 

And just to be clear, this fix _will_ use the callback method for older versions of metamask, but will fall back to polling if this isn't possible? Where exactly is this decision made in the code?

franckc:
Thanks for working on this ! Will review shortly. In the meantime, do you mind looking at those build failures ? it's just the linter complaining.

src/services/contract-service.js
sparrowDom: I just reworked this part since I wasn't happy with canceling functionality that worked in some cases. Working on a revamp of this. My idea is: 
- keep the on("confirmation") logic in 
- also start pooling the fallback function
- if we receive the event in on("confirmation") cancel the fallback function pooling, since we don't need it

franckc:
Adding @tyleryasaka for an extra pair of eyes on this origin-js core change...

sparrowDom:
@franckc @micahalcorn @tyleryasaka  did a bit of refactoring:
- send & deploy events are handled by the same function
- on("confirmation") event is still registered & alongside fallback pooling function is spawned
- if on("confirmation") is received the pooling is stopped since it is not needed

sparrowDom:
@franckc @micahalcorn @tyleryasaka @wanderingstan have updated the PR description, to help understand what the code does. The change is complex, since it handles a lot of the cases. But on the upside it solves both DApp issues mentioned in description. 

I have done testing with Metamask 4.12.0 on Rinkeby and it worked for ListingCreation, both profile saving cases (contractDeployment,  contract transaction)

I will do more testing tomorrow, and would really appreciate feedback. 🙏

DanielVF:
Hah. I'm the king of hating dependencies. (the majorhammer.com site has zero dependencies outside the language and the database.)

On the other hand, underscore/lodash is comparatively small, ultra-stable, and can make for much cleaner code. The code here is cleaner than it would be otherwise. Underscore/lodash often one of the few libraries I us -it has the least downside of all possible dependencies.

That said, I think our project's current vibe has been to use the advanced javascript features that babel gives, but not more than that. This has made for a fairly consistent javascript codebase. I think staying consistent and unsurprising is probably worth it. ( That said, I feel like I'm hitting a favorite child by saying so. :( )

DanielVF:
That was fast!

sparrowDom:
@DanielVF don't worry I am no super hero coder 🤸‍♂️  Was just finishing with reimplementation of the 2 lodash functions that we require. I feel like in college again when implementing basic mapping functions :)

Otherwise I totally agree with your point. Maybe a good middle ground would be to implement the mapping functions that we need and put them in utils. This way we should have best of both worlds: clean code in contract-service and no additional dependancies.

tyleryasaka:
I'll leave it to @franckc to determine if/when to merge this, but LGTM

franckc:
Thanks @sparrowDom for all your work ! Let's sync up on Mon and decide next steps.
We've got a demo coming up on Monday so I don't want to deploy risky changes until then... :)

sparrowDom:
I don't fully agree with that for the following reasons: 
- Metamask is totally DApp environment specific dependancy and for implementing one way for non 4.12 users an another for others would mean we need to either: 
  - inform `origin-js` what version of Metamask is being used
  - extend `origin-js` api so that DApp can issues different function calls for different Metamask versions
(both of these ugly up the code and documentation)
- Displaying warning on specific Metamask versions + supporting different code paths for them is IMHO bad for maintainability. This is a good way to introduce regressions. Also for each release instead of supporting the latest Metamask version, our QA needs to test for regressions with different Metamasks in environment. 
- Metamask 4.12.0 is so broken that most DApps won't function with it. Should we really take the burden + extra time/development costs for their mistakes.

cuongdo:
I agree that separate code paths makes regressions more likely. However, I don't think that allowing DApps built on top of Origin.js to be broken for > 1 week with the latest available MetaMask is the right experience for users either. We will sometimes need to make things less clean to deal with the reality of immature infrastructure. We should provide a better UX than most DApps.

I'm not clear on what the best set of tradeoffs here is, or whether we just need to make progress on a mobile wallet.

sparrowDom:
It would probably be beneficial for other to weigh in (@micahalcorn @wanderingstan @franckc @tyleryasaka). So the current options are: 
1. Show a warning when users are using non fully supported version of Metamask. Keep implemented workaround to mitigate Metamask errors. 
2. Show a warning that no write operations are permitted with specific versions of Metamask. Prompt users to upgrade.

sparrowDom:
I guess we have achieved a stalemate and both sides have good arguments.

If I try to sum up @cuongdo and @DanielVF lean more towards merging functionality from this PR and @tyleryasaka and me more against doing it. Should we flip a coin or shake the magic 8️⃣ 🏀 ? 
🤷‍♂️

micahalcorn:
🎱

sparrowDom:
Aaaa so there is an 🎱emoticon! 🎉

micahalcorn:
I can’t let that be my only contribution here. 😬

Perhaps the course of action depends on the nature of the bug. As Daniel points out, there May be other reasons for adding the polling code, in particular. I don’t think that there is any question that the mobile wallet is a priority for Q4. I also don’t think that we should go out of our way to support anything but the current version of MetaMask. So I guess my opinion is a bit nuanced:

- merge this PR
- have a general DApp warning that we don’t fully support MetaMask (if it is the current provider)
- encourage use of our mobile app
- only further clutter origin-js when we feel like the fix is not just just a short-term patch for the then-current version of MetaMask

tyleryasaka:
I've stated my opinion but I also haven't looked at the specifics too closely here - do what you gotta do, I won't be heartbroken either way. 😃

sparrowDom:
Closing since work is continued here: https://github.com/OriginProtocol/origin/pull/803

